PR #44, opened 2025-06-13
Relocated backend management to the engine.

shsym:
I have changed the system architecture and CLI's functionality based on our previous discussion, specifically:
- Used in the controller, the CLI can start the engine, with the engine-manager API server.
- Used in the backend, the CLI can start backend instances, which will register themselves to the engine.

@ingim This is a separate version of engine-manager, before we merge it with the engine-side server logic.


backend/backend-python/qwen3_cli.py
shsym, line 1: Ignore this; just a test file.

cli/src/backend.rs
```diff
@@ -0,0 +1,590 @@
...
+        /// Backend ID
+        backend_id: String,
+
+        /// URL of the management service
+        #[arg(long, default_value = network::DEFAULT_MANAGEMENT_URL)]
+        management_service_url: String,
+    },
+
+    /// Start a new backend process
+    Start {
```
ingim, line 33: This command is only to start the backend instance on the same host, right?

shsym: Yes, that's what we discussed before—no remote launch for backends (from CLI).

engine/src/service.rs
```diff
@@ -1,14 +1,14 @@
...
-use std::sync::OnceLock;
+use std::sync::{Arc, Mutex, OnceLock};
 use thiserror::Error;
 use tokio::sync::mpsc::{UnboundedSender, unbounded_channel};
 use tokio::task;
 
 // Common driver routines
 
-static SERVICE_DISPATCHER: OnceLock<ServiceDispatcher> = OnceLock::new();
+static SERVICE_DISPATCHER: OnceLock<Arc<Mutex<ServiceDispatcher>>> = OnceLock::new();
```
ingim, line 11: The ServiceDispatcher should be lock-free as it handles all API calls within the engine (whether from inferlets or servers) and having mutex will create a bottleneck. 

Let's consider other possible ways.

shsym: Is the lock called every time it serves API? Is it only for service id related APIs (not locking the entire API serving logic)? If it is for service id related tasks, I don't think it will cause too much overhead. We can think of more finer-grained locks once we see the bottleneck. For example, we can use RWLock so that almost cases will simply share the reader lock without causing much contention.

engine/src/service.rs
```diff
@@ -86,10 +123,10 @@ impl Controller {
     }
 
     pub fn install(self) {
-        let dispatcher = ServiceDispatcher {
+        let dispatcher = Arc::new(Mutex::new(ServiceDispatcher {
             maps: self.maps,
```
ingim, line 127: Instead of Arc-Mutexing the ServiceDispatcher,

We can use DashMap for maps, and some lock-free vector for channels, e.g., something like this (https://github.com/ibraheemdev/boxcar)

shsym: Boxcar is append only, while service dispatcher might require removing service. For now, I will suggest using RWLock.

engine/src/main.rs
```diff
@@ -199,149 +172,32 @@ async fn main() -> anyhow::Result<()> {
...
-    }
-
-    client
-        .launch_server_instance(&program_hash, port as u32)
-        .await?;
-
-    Ok(())
-}
-
-async fn dummy_client(program_name: String) -> anyhow::Result<()> {
```
ingim, line 276: For quick debugging purposes, I prefer to have the client loading option when launching the engine. Can we have these snippets back?

shsym: We cannot do it now since the backend should be connected first then the application can be launched (previously they were all hardcoded).
As we can launch controller (engine), backend, and applications with CLI, I would suggest using them for debugging purposes as well. It would be just a script having CLI calls ;)

engine/src/service.rs
ingim: Every API calls from (concurrent) inferlets will invoke the service dispatcher.

shsym: Yes, and if they can get the lock with reader permission that can be shared across inferlets, there will be no much contention.

engine/src/service.rs
ingim: It is okay to use an append-only vector since it only stores a channel. To remove a service, we can simply delete it from the service ids after cleaning them up.

after all, the point of ServiceDispatcher is to minimize the overhead of sync as much as possible. RWLock kinda defeats its purpose.

shsym: I see your point. Maybe a more fundamental way to address this is to use HashMap such as DashMap as you suggested above.

ingim: Some locking is unavoidable like managing service IDs with maps.

I’m fine using RwLock<HashMap> or DashMap for `maps`. Because the map doesn’t store the channels directly to avoid repeated lookups.

API-heavy services (e.g., [pie l4m.rs](https://github.com/pie-project/pie/blob/main/engine/src/bindings/l4m.rs#L14)) cache service IDs to skip hashmap lookup on every call.

Service IDs directly index into the routing `channels` vector, so as long as we don't have locks on the channel vector, dispatch overhead is minimal.

PR #54, opened 2025-09-10
Enhance linting workflow with Black and Pyright

shsym:
Updated linting steps to include Black and Pyright for predefined and changed files.

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI to run Python linting (Pylint) and type checking (Pyright) on a targeted set of changed and key files for faster, focused feedback.
  * Improved log messages for code formatting to be clearer.
  * Workflow now gracefully skips checks when no relevant files are detected, reducing noise.
  * Provides explicit visibility into which files are being analyzed during linting and type checks.

### DIFF
--- a/.github/workflows/pylint_v1.yml
+++ b/.github/workflows/pylint_v1.yml
@@ -51,5 +51,102 @@ jobs:
             echo "No Python files changed under backend/backend-python."
             exit 0
           fi
-          echo "Linting: $FILES"
+          echo "Linting with Black: $FILES"
           uvx black --check $FILES
+
+      - name: Run Pylint on predefined files + changed files outside those dirs
+        shell: bash
+        env:
+          # keep uv offline/safe as in your scripts
+          UV_NO_SYNC: 1
+          UV_OFFLINE: 1
+          PYTHONPATH: backend/backend-python
+        run: |
+          set -euo pipefail
+          # allow globs that don't match to disappear
+          shopt -s nullglob
+
+          # Predefined targets (expand if files exist)
+          PREDEF=(backend/backend-python/config/*.py backend/backend-python/model/*.py)
+
+          # Changed files space-separated from previous step
+          CHANGED_STR="${{ steps.changes.outputs.files }}"
+          CHANGED_LIST=()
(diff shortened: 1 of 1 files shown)


shsym:

---

## Correction: the collision this PR fixes is not reachable today

I justified this change by saying two driver groups in one process is "every
data-parallel deployment, not a hypothetical." That is wrong, and since it was
the stated reason for merging, it should not stand.

`worker/src/preflight.rs` refuses it outright:

```rust
let num_groups = world_size / tp_degree;
if num_groups > 1 {
    anyhow::bail!(
        "model.driver.device lists {world_size} devices with \
         tensor_parallel_size = {tp_degree}, which asks for \
         {num_groups} data-parallel replicas in one engine. A worker \
         serves one replica: run {num_groups} workers, each with \
         {tp_degree} device(s), and let the gateway spread requests \
         over them."
    );
}
```

There is one `calculate_topology` call site, one `load_model_drivers` call per
process, and `preflight.rs` has a test asserting `calculate_topology(4, 2)`
errors. Data parallelism runs as **separate processes**, and `TpPhaseWatch` is a
per-process singleton — so two groups never share it. I read the loop
`for (group_idx, group) in topology.iter().enumerate()` and inferred it could
run more than once without checking what feeds it.

## What that means for this change

Being straight about the rest of it too:

- The **duplicate reporting for a four-rank group** and the **process-wide
  watermark** were not pre-existing bugs I found. Both were regressions
  introduced by this PR's own first commit — the per-slot restart and the shared
  counter came in with the group loop — and then fixed later in the same PR.
  Before it, `fire_disagreements` ran one loop with one watermark, which is
  correct for one group.
- The **missed bind site** is the same story: with rank-indexed slots the leader
  was fine; it only needed binding because this change made slots claimable.

So this fixes no live defect. What it leaves behind is an invariant that is now
enforced rather than described, a report that can name a group, and a test that
pins six properties. That is worth having when a second group becomes possible,
and it costs one relaxed load on the publish path — but it is hardening, not a
bug fix, and I should have said so.

shsym/tts-pie#42 stays closed: the addressing it asks for is implemented. Anyone
weighing a revert should know the collision is currently unreachable.

Issue #60, opened 2025-09-15
Separate scheduler for heartbeats


ingim:
The current implementation runs both GPU operations and heartbeat management on a single thread. During the first Pie initialization, the GPU loop may block for longer than the configured heartbeat timeout. This can result in missed heartbeats and unintended failures.

Introduce a dedicated thread for heartbeat management that runs independently of GPU operations.

PR #77, opened 2025-09-30
Split the heartbeat response into a separate thread

zyma98:
The FlashInfer kernels sometimes need JIT compilation, which can take a long time and block the worker thread, causing timeout in the heartbeat response. By splitting the heartbeat response into a separate thread, we can avoid the timeout issue.


zyma98:
Fixed Issue #60

PR #249, opened 2026-01-30
Unify local and registry inferlet launching

zyma98:
Implement the change proposed in https://github.com/pie-project/pie/discussions/205#discussioncomment-15623480 Packet Search Path

Now, when launching an instance, the engine will first search for the program among client-uploaded programs. If not found, it will fall back to searching the registry. The launching instance from registry function still forces searching only the registry.

It is now required that the client passes the `Pie.toml` manifest file when uploading a local inferlet to the engine. This will be important for future dependency resolution for dynamic linking.

This PR address an inefficiency in the previous code, where downloaded inferlets from the registry stored in the registry cache were copied into the directory that containing other inferlets uploaded from clients. Now, there is no more copying.

This PR also identified and fixed three potential Tokio thread starvation problems:
1. Previously, disk operations and Wasm compiling are all done by the single runtime handler Tokio thread. These jobs are now moved to client handler threads in the server, where we have one Tokio thread per client connection.
2. Disk operations were using blocking `std::fs` functions. They are now asynchronous `tokio::fs` operations.
3. Wasm compiling could take a long time that starves the Tokio async runtime. The compilation is now moved to a blocking thread and the Tokio async thread `await` on this blocking thread.

### DIFF
--- a/client/rust/src/client.rs
+++ b/client/rust/src/client.rs
@@ -409,14 +409,14 @@ impl Client {
     /// - Without version (defaults to `latest`): `std/text-completion` or `text-completion`
     pub async fn launch_instance(
         &self,
-        inferlet: &str,
+        inferlet: String,
         arguments: Vec<String>,
         detached: bool,
     ) -> Result<Instance> {
         let corr_id_guard = self.inner.corr_id_pool.acquire().await?;
         let msg = ClientMessage::LaunchInstance {
             corr_id: *corr_id_guard,
-            inferlet: inferlet.to_string(),
+            inferlet,
             arguments,
             detached,
         };


client/python/src/pie_client/client.py
```diff
@@ -385,9 +385,23 @@ async def query(self, subject: str, record: str) -> tuple[bool, str]:
         msg = {"type": "query", "subject": subject, "record": record}
         return await self._send_msg_and_wait(msg)
 
-    async def program_exists(self, program_hash: str) -> bool:
-        """Check if a program with the given hash exists on the server."""
-        successful, result = await self.query("program_exists", program_hash)
+    async def program_exists(self, inferlet: str, hash: str | None = None) -> bool:
```
ingim, line 388: I like how it can optioanlly check for hash consistency

zyma98: This is actually pretty crucial. Otherwise, after an inferlet is re-compiled locally, the engine would confuse the existing old one with the new one.

ingim:
Great work @zyma98 !

PR #250, opened 2026-01-31
Add dependency parsing to engine

zyma98:
Implement the change proposed in https://github.com/pie-project/pie/discussions/205#discussioncomment-15623480 Dependency Specification

The engine now reads an optional `dependencies` field under the `[package]` section of the inferlet manifest TOML file. When loading an inferlet to the runtime, the engine ensures that its dependencies have already been loaded.

The engine runtime now uses the pair of `(wasm_hash, toml_hash)` to identify a loaded inferlet (or library). This is necessary to detect changes in manifest when the client launches from a local inferlet with only changes in manifest but not the Wasm binary. If we don't include the `toml_hash` part, the engine may use stale dependencies of an already loaded inferlet.

To make recursive dependency resolution efficient, the engine maintains the following invariant: If an inferlet or a library is loaded in the runtime, then all of its dependencies must have been loaded. With this invariant, as long as an immediate dependency is already loaded, we need not to recursively check the recursive dependencies of this already loaded dependency.

## Summary by CodeRabbit

* **New Features**
  * Install flow replaces upload: programs are installed from WASM+manifest paths.
  * CLI: new --manifest option for path-based runs (manifest required with local path).
  * Automatic dependency resolution and recursive loading from the registry.

* **Improvements**
  * Program existence checks now validate both WASM and manifest.
  * User messages reflect manifest-driven install/launch workflow and clearer install logs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/benches/tput.py
+++ b/benches/tput.py
@@ -47,10 +47,10 @@ async def run_benchmark(args):
     async with PieClient(args.server) as client:
         await client.authenticate("benchmark-user")
 
-        # 3. Upload program (check both name and hashes match)
+        # 3. Install program (check both name and hashes match)
         if not await client.program_exists(inferlet_name, wasm_path, manifest_path):
-            print("Uploading program...")
-            await client.upload_program(wasm_path, manifest_path)
+            print("Installing program...")
+            await client.install_program(wasm_path, manifest_path)
         else:
             print("Program already exists on server.")
 
--- a/client/javascript/src/index.js
+++ b/client/javascript/src/index.js
@@ -346,12 +346,12 @@ export class PieClient {
     }
 
     /**
-     * Uploads a program to the server in chunks.
+     * Installs a program to the server in chunks.
      * @param {string} wasmPath Path to the WASM binary file (Node.js only).
      * @param {string} manifestPath Path to the manifest TOML file (Node.js only).
      * @returns {Promise<void>}
      */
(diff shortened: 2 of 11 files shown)


zyma98:
There is a misunderstanding and let me try to clarify. For now I'm ditching the link-time polymorphism design, as I haven't find a compelling use case for it nor a good user interface. The dependencies specified in the TOML should not be interpreted as a generic interface, but the absolute name of the dependency library.

Sorry that I confused you with the ELF binary example, but my point is that two programs are different if the dependencies are different. A program logically is formed by the application binary plus all of its dependencies. If the dependency changes, the program changes, as if the program is statically linked. Since dependencies are specified in the TOML, we just use the TOML hash to track dependency changes.

In the past, I've talked a lot about link-time polymorphism, but we are not doing it here. I should have clarified it earlier, my bad.

ingim:
suppose a user has uploaded two inferlets, `foo` and `bar`, where `foo` depends on `bar`. If `bar` is overwritten without changing its version, `foo` would need to be re-linked.

Given that `upload_program` can trigger state changes in other existing inferlets, would it make sense to rename it to `install_program`? "Install" might better convey that there could be broader effects beyond just uploading.

zyma98:
Sounds good. I'll change the name to `install_program`.

zyma98:
New changes are ready for review.

PR #252, opened 2026-02-01
Support dynamic linking

zyma98:
Support Wasm component dynamic linking via proxy forwarding from host. This PR implements the change proposed in https://github.com/pie-project/pie/discussions/205#discussioncomment-15623480

We need to extend `bakery inferlet publish` so that the dependency specification in included when publishing.

For now, to test the code, please unzip the following attached zip file and move it to `.pie/cache/registry/zyma98`. Then, run the following command to run with dynamic linking:

```
pie-client submit zyma98/text-completion-inferlib -- --prompt Hello
```

[zyma98.zip](https://github.com/user-attachments/files/24997338/zyma98.zip)

### DIFF
--- a/runtime/src/instance.rs
+++ b/runtime/src/instance.rs
@@ -14,7 +14,7 @@ use std::task::{Context, Poll};
 use tokio::io::AsyncWrite;
 use tokio::sync::Notify;
 use uuid::Uuid;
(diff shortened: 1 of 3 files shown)


ingim:
LGTM. Do you want to merge this before Bakery has dependency support, or wait and merge afterwards?

PR #260, opened 2026-02-08
fix(worker): remove <|im_start|> from Qwen3 stop tokens

shsym:
## Summary

- Remove `<|im_start|>` from `Qwen3Template.stop_tokens` in `chat_templates.py`
- `<|im_start|>` is a message boundary marker, not a generation stop signal
- This caused beam search to return empty responses when a beam produced `<|im_start|>` as its first token, immediately triggering the stop condition

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Qwen3 model text generation stopping behavior.


shsym:
I merge this as a one-line quick fix.

PR #315, opened 2026-04-15
fix(sdk/python): pass actual_pages to forward.kv_cache

shsym:
WIT interface requires actual_pages as the 4th arg, but Python binding still had the old 2-arg signature, causing TypeError at runtime for any inferlet using generate/decode paths (beam search, generate, decode_step_dist).

Default actual_pages to len(kv_page_ptrs) so the 3 existing call sites in context.py (decode, decode_dist, decode_step_dist) keep working — Python's decode path never pre-allocates trailing pages, unlike the Rust SDK.


shsym:
Merge: non-main branch.

PR #316, opened 2026-04-16
feat(server): let clients choose the bind host for launch_server_instance

shsym:
## Motivation

The server inferlet binding is hardcoded to \`127.0.0.1\` in
\`runtime::launch_server_instance\`. That's the right default — a loopback
HTTP server owned by a trusted parent process — but it blocks a real
use case: running \`pie http\` inside a Docker container on bridge
networking and exposing the server via \`-p 8095:8095\`. Under bridge
networking Docker can only port-forward to the container's \`0.0.0.0\`,
so a \`127.0.0.1\` bind is unreachable from the host.

The immediate driver is the \`pieclaw/docker/inferlet-runner\` image
(just merged) — it's currently forced onto \`--network host\` purely
because of this constraint, which leaks host-port collisions between
multiple runner containers (every new one has to pick its own engine
WS port to avoid stepping on 8080).

## What changes

Optional \`host\` field threaded end-to-end through the
\`launch_server_instance\` RPC:

- **\`client/rust/src/message.rs\`** — new \`host: Option<String>\` on
  \`LaunchServerInstance\`, \`#[serde(default)]\` so older clients still
  deserialize.
- **\`runtime/src/server.rs\`** — destructure + forward through
  \`handle_launch_server_instance\` into the runtime \`Command\`.
- **\`runtime/src/runtime.rs\`** — \`Command::LaunchServerInstance\` grows
  a \`host: Option<String>\` variant field. \`launch_server_instance\`
  parses the host string into an \`IpAddr\` (errors with a clear message
  on garbage input) and uses it in the \`SocketAddr\`. \`None\` falls
  through to \`127.0.0.1\`.
- **\`client/python/src/pie_client/client.py\`** — \`host: str | None = None\`
  kwarg; only included in the outgoing msgpack when set, preserving
  wire-compat with older runtimes.
- **\`pie/src/pie_cli/http.py\`** — new \`--host\` typer option defaulting
  to \`127.0.0.1\`. The \"Listening on...\" panel + message now echo the
  chosen host.

## Wire compatibility

- **Old client → new runtime**: no \`host\` in payload → \`serde(default)\`
  gives \`None\` → default \`127.0.0.1\`. ✅
- **New client, default host → old runtime**: Python helper omits the
  key when host == 127.0.0.1, so the msgpack wire bytes are identical
  to the old path. ✅
- **New client, \`--host 0.0.0.0\` → old runtime**: the old runtime fails
  msgpack deserialization with an unknown field. Expected & acceptable —
  only surfaces when the operator opts in.

## Test plan

- [x] \`cargo check\` passes for \`runtime/\` + \`client/rust/\` on the
      pieclaw-native-dev image.
- [x] \`python3 -m py_compile\` passes for the changed Python files.
- [ ] Smoke test on ECL: \`pie http --path ... --port 18099 --host 0.0.0.0\`
      binds \`0.0.0.0:18099\` (and reaches from host); default binds
      \`127.0.0.1:18099\` as before.

## Follow-up

Once merged + bumped in \`pie-vllm\`, the \`pieclaw/docker/inferlet-runner\`
image drops the \`ENGINE_PORT=PORT+10000\` workaround and switches from
\`--network host\` to \`--network bridge -p EXT:INT\`, which gives each
runner container its own netns. Tracking in shsym/pieclaw#12.

### DIFF
--- a/client/python/src/pie_client/client.py
+++ b/client/python/src/pie_client/client.py
@@ -646,6 +646,7 @@ class PieClient:
         program_hash: str,
         port: int,
         arguments: list[str] | None = None,
+        host: str | None = None,
     ) -> None:
         """
         Launch a server inferlet that listens on a specific port.
@@ -657,14 +658,18 @@ class PieClient:
         :param program_hash: The hash of the uploaded program.
         :param port: The TCP port to listen on.
         :param arguments: Command-line arguments to pass to the inferlet.
+        :param host: IP to bind on (e.g. ``"0.0.0.0"``). ``None`` keeps the
+            engine's historical default of ``127.0.0.1``.
         :raises Exception: If launch fails.
         """
-        msg = {
+        msg: dict = {
             "type": "launch_server_instance",
             "port": port,
             "inferlet": program_hash,
             "arguments": arguments or [],
         }
+        if host is not None:
+            msg["host"] = host
         successful, result = await self._send_msg_and_wait(msg)
         if not successful:
             raise Exception(f"Failed to launch server instance: {result}")
--- a/client/rust/src/message.rs
+++ b/client/rust/src/message.rs
@@ -108,6 +108,10 @@ pub enum ClientMessage {
         port: u32,
         inferlet: String,
         arguments: Vec<String>,
+        /// IP to bind the server inferlet on.  Omitted by older clients → defaults to
+        /// `127.0.0.1` runtime-side (see `server::handle_launch_server_instance`).
+        #[serde(default)]
+        host: Option<String>,
     },
 
     #[serde(rename = "signal_instance")]
--- a/pie/src/pie_cli/http.py
+++ b/pie/src/pie_cli/http.py
@@ -27,6 +27,13 @@ def http(
     port: int = typer.Option(
         ..., "--port", help="TCP port for the server to listen on"
     ),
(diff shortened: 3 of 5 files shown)


shsym:
### In-container build + partial smoke test

On `sslee0cs/pie-vllm-engine:pieclaw-native-dev`:

- `maturin develop --release` rebuilds the runtime cleanly (17 pre-existing warnings, 0 errors, 1m 08s) and reinstalls `pie-server-0.2.5` into `/opt/pie/pie/.venv`.
- `pie http --help` now shows the `--host` flag with help text.
- `pie http --path … --port 18099 --host 0.0.0.0 --config … --dummy` prints the new `Bind           0.0.0.0:18099` panel — confirms CLI → client → runtime plumbing through `_launch_server_inferlet` and `launch_server_instance(host=host)`.
- Default call (`pie http` without `--host`) continues to print `Bind           127.0.0.1:18099` — old behaviour preserved.

The bind itself wasn't exercised in this container (no nvidia runtime, engine init failed pre-bind with `Can't initialize NVML`). End-to-end verification on a GPU-enabled host is pending — plan is to fold it into the follow-up `pieclaw/docker/inferlet-runner` PR that drops `--network host` once this lands in `pieclaw/home`.

PR #322, opened 2026-04-24
fix(runtime): propagate fire_batch Python errors instead of silent empty completions

shsym:
## Summary

When the Python worker raises inside `fire_batch` (`pie/src/pie/manager.py:1418-1425` catches the exception and sends `{"error": str(e), "results": []}` on the side-channel), the Rust `BatchedForwardPassResponse` previously only had a `results` field, so `rmp_serde` silently discarded `error`. Every request in the failed batch took the no-response path in `execute_forward_pass_batch`:

- Multi-step: sent `ForwardPassResponse { tokens: [], dists: [] }` to WASM
- Non-multi-step: dropped `resp_tx`, oneshot gave the SDK `Canceled`

The inferlet SDK treats empty tokens as _\"engine returned no tokens — final exit\"_ and breaks out of `generate()`. The client received HTTP 200 with `usage.completion_tokens=0` and `finish_reason=\"stop\"`. An upstream `AssertionError` at `execute_step` thus looked identical to a normal short completion on the wire, masking production incidents.

Observed downstream (pie-hermes task #48): at N=32 on Qwen2.5-72B-AWQ, 54% of requests reported `completion_tokens=0` with zero HTTP errors while the engine log showed `[ERROR] execute_step failed: AssertionError` and `[FIRE-BATCH-SOCK Error]`.

## Changes

1. Add `error: Option<String>` to `BatchedForwardPassResponse` with `#[serde(default)]`. Wire-compatible with existing success payloads (absent key → `None`).
2. On a fire_batch error, propagate via `terminate_instance_with_exception` per request so the client receives a proper error instead of a successful-looking empty completion.
3. Two deserialization unit tests covering the error payload and the backward-compatible success payload.

## Scope

This patch makes the failure **observable**, not **prevented**. The underlying AssertionError that triggers fire_batch failure at high concurrency needs a separate fix once the error surface is clear.

## Test plan

- [x] `cargo test --lib request::tests::batched` — 2 new tests pass
- [x] `cargo check --manifest-path runtime/Cargo.toml` — clean
- [ ] End-to-end: rebuild pie on pod, rerun the N=32 bench from pie-hermes/task #48, confirm clients now see a 5xx instead of HTTP 200 + empty completion (in flight)

## Related

- pie-hermes task #48 (reporter)
- pie-hermes PR #14 (wedge fix that unblocked observation of this bug — without that fix the requests hung at 300s so this codepath never fired)

## Release Notes

* **New Features**
  * Added configurable server host binding (default `127.0.0.1`).
  * Introduced vLLM backend as an alternative inference engine option.
  * Added server-side chat formatting with HuggingFace template support.
  * Exposed model-level generation defaults (temperature, top-p, top-k, min-p, repetition-penalty).
  * Added token sampling penalties (repetition, frequency, presence) for fine-grained control.
  * Enabled multi-step decode operations per execution call.
  * Improved batch scheduling and request merging for throughput optimization.

* **SDK Enhancements**
  * Rust SDK: streaming token execution and KV cache synchronous export.
  * Python SDK: extended forward pass with actual KV page tracking.

### DIFF
--- a/runtime/src/model.rs
+++ b/runtime/src/model.rs
@@ -986,6 +986,30 @@ impl Model {
 
         match result {
             Ok(batch_resp) => {
+                // Python raised inside fire_batch: propagate as per-request
+                // instance termination rather than silently dropping into the
+                // "no response → empty tokens" path that used to mask errors
+                // as HTTP 200 + completion_tokens=0 (task #48).
+                if let Some(err_msg) = batch_resp.error.as_ref() {
+                    eprintln!(
+                        "[Error] fire_batch returned error for batch={} reqs={}: {}",
+                        trace_batch_id, batch_size, err_msg
+                    );
+                    for (fp_req, resp_tx) in requests {
+                        // Drop streaming sender so WASM sees the stream end
+                        // before the termination notification arrives.
+                        drop(fp_req.token_stream_tx);
+                        drop(resp_tx);
+                        if let Some(inst_id) = fp_req.inst_id {
+                            terminate_instance_with_exception(
+                                inst_id,
+                                format!("fire_batch failed: {}", err_msg),
+                            );
+                        }
+                    }
+                    return;
+                }
+
                 // Collect all continuations FIRST, then send them in bulk.
                 // This prevents tokio from interleaving the scheduler loop
                 // between individual sends, which causes batch fragmentation
--- a/runtime/src/model/request.rs
+++ b/runtime/src/model/request.rs
@@ -595,10 +595,21 @@ impl Default for BatchedForwardPassRequest {
 }
 
 /// Batched forward pass response from Python via pycrust.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+///
+/// `error` is populated by the Python fire_batch wrapper when the batch
+/// raised an exception (see pie/src/pie/manager.py). Prior to its addition,
+/// rmp_serde silently discarded the error string because the Rust struct
+/// only had `results`; failed batches manifested on the client as HTTP 200
+/// with completion_tokens=0 (task #48). Callers MUST check `error` before
+/// treating an empty `results` as successful completion.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct BatchedForwardPassResponse {
-    /// Results indexed by request order in the batch.
+    /// Results indexed by request order in the batch. Empty when `error` is set.
+    #[serde(default)]
     pub results: Vec<ForwardPassResponse>,
+    /// Python-side exception string from fire_batch (None on success).
+    #[serde(default)]
+    pub error: Option<String>,
 }
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


shsym:
## E2E verification (A100-80GB, Qwen2.5-72B-AWQ, N=32 S1, patched _pie.so deployed)

### What the patch DOES solve (Phase 1 — observability)

Pre-patch, the upstream `AssertionError` at `execute_step` was completely silent to operators and to the runtime-side log. Post-patch:

```
[ERROR] execute_step failed:
Traceback (most recent call last):
  File ".../pie_worker/vllm_runtime.py", line 1004, in execute_step
    result = self.vllm_worker.execute_model(scheduler_output)
  ...
  File ".../vllm/v1/worker/gpu_model_runner.py", line 1204, in _update_states
    self.input_batch.add_request(request)
  AssertionError

[FIRE-BATCH-SOCK Error] AssertionError
...
[Error] fire_batch returned error for batch=12 reqs=1: AssertionError
[Error] fire_batch returned error for batch=13 reqs=1: AssertionError
...
```

Counts from one N=32 bench window:

| signal | count |
|---|---|
| `execute_step failed` (Python traceback) | 10 |
| `[FIRE-BATCH-SOCK Error]` | 10 |
| `fire_batch returned error` (new — this PR) | **157** |
| `hyper::Error` | 0 |

The Python exception string now surfaces all the way to the Rust log via the new `error: Option<String>` field — operators can grep the exception text without having to correlate side-channel logs.

### What this patch does NOT yet solve

HTTP-surface behavior is unchanged in this bench: 320/320 `ok`, 236/320 empty `content`, zero HTTP 5xx. Root cause of the residual behavior traced during verification:

`terminate_instance_with_exception` dispatches `runtime::Command::TerminateInstance`. The handler at `pie/runtime/src/runtime.rs:1050-1053` only looks up the instance in `running_instances` / `finished_instances`. **Per-HTTP-request WASM futures spawned in `handle_server_request` are NOT tracked in any DashMap** — the `inst_id` we generate at runtime.rs:1151 is just a tag used by `model::cleanup_instance` — so `terminate_instance` is a no-op for them. The WASM task therefore runs to completion, sees zero tokens from `decode_n` (which the SDK at `context.rs:1023` treats as natural final-exit), builds an empty-content response, and returns HTTP 200.

### Shape of a follow-up (pie-hermes task #52 / Phase 2)

Full propagation needs one of:

1. Extend `terminate_instance` to also look up `running_server_instances` — but that only tracks the top-level listener task, not the per-request spawn; would need a new per-request registry.
2. Change `resp_tx` from `oneshot::Sender<ForwardPassResponse>` to `oneshot::Sender<Result<ForwardPassResponse, String>>` and have the SDK's `decode_n` propagate `Err` up through `generate()` so the WASM handler emits a 5xx.
3. Have the SDK distinguish "oneshot Canceled before first token" (engine error) from "Ok(empty)" (natural stop) and error out on the former.

Recommend landing this PR as the observability step and treating full HTTP-status propagation as Phase 2.

shsym:
## Phase 2 E2E verification (pod 275uf3bfgvunpi, A100-80GB, Qwen2.5-72B-AWQ)

New commit `74117828` adds `error: Option<String>` on `ForwardPassResult` and makes `get_tokens` trap the guest with an anyhow error when the oneshot is Canceled before any token was streamed. Result at N=32 S1 (320 requests total):

| | Phase 1 only | **Phase 1 + Phase 2** |
|---|---|---|
| HTTP 200 total | 320 / 320 | 77 / 320 |
| HTTP 200 with **empty** `content` (silent failure) | **236 / 320** | **0 / 320** |
| HTTP 200 with real content | 84 / 320 | 77 / 320 |
| HTTP errors (client visible) | 0 / 320 | 243 / 320 |

Client-side errors are all `Remote end closed connection without response` — the hyper-side surfacing of:

```
error: error while executing at wasm backtrace:
error: hyper::Error(User(Service), guest never invoked `response-outparam::set` method
```

Which is what we want: the guest instance traps on the Canceled oneshot, hyper reports "guest never set response-outparam" and closes the connection. Clients now see an error instead of a successful-looking empty completion. Retry logic / alerting on error-rate now works correctly.

Note the remaining 77/320 (24 %) with real content are genuine successes — the upstream `AssertionError` at vLLM's `input_batch.add_request` does not fire on every batch at this concurrency, only on some (~75 % of the time in this run).

The underlying vLLM assertion at N≥32 is a scheduler issue orthogonal to this PR (tracked in pie-hermes task #52); this PR's goal is to make such failures observable rather than silent, which Phase 2 completes.

## Commits in this PR

1. `2026e4da` fix(runtime): propagate fire_batch Python errors — engine-log surface (Phase 1)
2. `74117828` fix(runtime): surface Canceled-before-first-token as a guest trap — HTTP 5xx surface (Phase 2)

## Residual polish (not in this PR)

The client sees "Remote end closed connection without response" rather than a proper `500 Internal Server Error` body. A nicer end-state would have hyper translate the trap into a structured 5xx response with the Python exception string. Would need a larger change in `handle_server_request` (runtime.rs). Leaving for follow-up.

PR #330, opened 2026-04-30
fix(arch-resolution): own HF→pie arch translation in Python; Rust dispatches on pie name only (closes #328)

### LINKED ISSUE #328: Qwen3 .with_reasoning(): `thinking` field stays empty when `</think>` is not reached

shsym:
## TL;DR

- **Pie internal name → `Instruct` impl** is already what Rust dispatches on; we just stopped the vllm/sgl drivers from forwarding HF class names like `Qwen3ForCausalLM` and route them through the same `HF_TO_PIE_ARCH` table the native driver always used.
- **The table stays exactly where it lives today** (`pie_driver/model/__init__.py`); we add `resolve()` next to it, and make the per-arch module imports lazy so callers reading the table don't trigger the `pie_kernels` JIT — no new module file introduced.
- **Two Rust unit tests pin the contract**: a resolved pie name (`"qwen3"`) enables thinking; a raw HF class name (`"Qwen3ForCausalLM"`) hits the silent fallback (regression test for #328). E2E reproducer flips from `{"thinking":"","text":"<think>…"}` to `{"thinking":"\nOkay, …","text":""}`.

Closes #328. Discussion of follow-ups (long-term home, the latent silent-fallback for `qwen3_5` / `phi3` / `mixtral` / `gemma4` / `dummy`, docs-drift) lives in #334.

---

## What was actually broken

`pie_driver_vllm/loader.py:212-220` and `pie_driver_sgl/loader.py:209` forwarded `architectures[0]` verbatim — `"Qwen3ForCausalLM"` for Qwen3-0.6B. Rust matched only `"qwen3"`, so vllm/sgl loads silently fell into the `_` arm with `has_thinking: false` (`NoopReasoningDecoder`, which always emits `Delta(empty)`). In the SDK pipeline that lets the chat decoder render token id 151667 as the literal string `"<think>"` and emit it as `Event::Text`, so reasoning content leaked into `text` while `thinking` stayed empty. The legacy `pie_driver` path was unaffected — its `loader.py:65-83` already translated via `HF_TO_PIE_ARCH`.

### 1. Single owner for HF → pie translation, in its existing location

`pie_driver/model/__init__.py` already had `HF_TO_PIE_ARCH` and the `register(..., hf_model_types=...)` populator. The PR adds a `resolve(hf_config) → str` function and `UnknownArchitectureError` in the same file, so callers stop reaching for `architectures[0]` themselves:

```python
# pie_driver/model/__init__.py — additions
def resolve(hf_config) -> str:
    """Resolve an HF config to the pie internal arch name.
    Future: rope_scaling-based llama2 vs llama3 disambiguation.
    """
    model_type = _attr_or_key(hf_config, "model_type", "")
    if model_type in HF_TO_PIE_ARCH:
        return HF_TO_PIE_ARCH[model_type]
    raise UnknownArchitectureError(...)
```

Why Python (not Rust) owns this: disambiguating `LlamaForCausalLM` between llama / llama2 / llama3 / llama3.1+ needs `rope_scaling.rope_type`; only the layer that holds the full HF config can do it.

### 2. Make the existing file cheap to import

The reason vllm/sgl couldn't reuse the existing table before: `pie_driver/model/__init__.py:109-110` did `from . import llama3, qwen2, …` at top-level, and every per-arch module under `pie_driver/model/` does `import pie_kernels as ops` (14 occurrences confirmed via grep), which JIT-compiles CUDA extensions at import time — heavy and irrelevant on the vllm/sgl path.

This PR switches `register()` to take a dotted module-path string and `get_module()` to `importlib.import_module` lazily on first dispatch:

```python
# Before
register("llama3", llama3, aliases=("l4ma",), hf_model_types=("llama",))
# After
register("llama3", "pie_driver.model.llama3", aliases=("l4ma",), hf_model_types=("llama",))
```

Same end state for the path that actually needs the module (the native engine still loads `qwen3.py` on its first `get_module("qwen3")` call); clean skip for vllm/sgl which never call `get_module()`. The "we avoid `pie_driver.hf_utils` because it transitively imports `pie_kernels`" comment at `pie_driver_vllm/loader.py:234-236` is now obsolete and dropped.

### 3. Migrate the four loader paths

| Driver | Now |
|---|---|
| `pie_driver_vllm/loader.py` | `from pie_driver.model import resolve` |
| `pie_driver_sgl/loader.py` | same |
| `pie_driver/loader.py` (native) | same (replaces its own `HF_TO_PIE_ARCH.get(...)` block) |
| `pie_driver/hf_utils.py::check_pie_compatibility` (CLI) | same (replaces its own lookup) |

`DriverCapabilities.arch_name`'s docstring updated: was *"first HF architecture string"*, now *"pie internal arch name (lowercase: 'qwen3', 'llama3', …); raw HF class names like 'Qwen3ForCausalLM' are NOT accepted on the Rust side."*

### 4. Rust regression tests

```
$ cargo test --lib model::instruct::tests::
running 2 tests
test model::instruct::tests::qwen3_pie_name_enables_thinking ... ok
test model::instruct::tests::hf_class_name_hits_fallback ... ok

test result: ok. 2 passed
```

`hf_class_name_hits_fallback` is the regression pin — if a future contributor adds an HF alias to `instruct::create()`, this test fails with a panic message pointing back at #328 and asking them to fix the driver instead.

## Diff shape

```
pie/src/pie/capabilities.py         | docstring rewrite
pie/src/pie_driver/hf_utils.py      | check_pie_compatibility() → resolve()
pie/src/pie_driver/loader.py        | replaces HF_TO_PIE_ARCH.get block with resolve()
pie/src/pie_driver/model/__init__.py| add resolve() + UnknownArchitectureError;
                                    | register() takes str path; get_module() lazy
pie/src/pie_driver_sgl/loader.py    | call resolve(); fix stale docstring; drop dead `arches=…`
pie/src/pie_driver_vllm/loader.py   | call resolve(); fix stale docstring; drop avoid-comment
runtime/src/model/instruct.rs       | doc comment + 2 unit tests
```

Net: ~+190 / -75 lines. **No new files added.** Original code structure preserved.

## E2E verification

Built `sslee0cs/pie-vllm-engine:deva-trial-fix328-v3` from this branch (cherry-picked onto `features/vllm-opt` for the cu129 stack), ran the canonical reproducer on ECL with `Qwen/Qwen3-0.6B`, prompt `"What is 2 + 2? Answer in one word."`, max_tokens=32, T=0.6, top_p=0.95, FlashInfer:

**Pre-fix (`deva-trial-fix3`):**
```json
{"thinking":"","text":"<think>\nOkay, the user is asking what 2 plus 2 is, and they want the answer in one word. Let me think. 2 plus"}
```

**Post-fix (this branch):**
```json
{"thinking":"\nOkay, the user is asking what 2 plus 2 is, and they want the answer in one word. Let me think. 2 plus","text":""}
```

Reasoning content lands in `thinking`; `text` stays empty until generation runs past `</think>`.

## Out of scope (covered in #334)

- Long-term home / namespace for the resolver.
- Latent silent-fallback for `qwen3_5`, `phi3`, `mixtral`, `gemma4`, `dummy` — registered in Python without Rust arms; same failure mode as #328 just for different arches.
- `website/docs/models.mdx` hand-maintained mirror.
- llama2 vs llama3 vs llama3.1+ disambiguation via `rope_scaling.rope_type` — referenced in `resolve()` docstring as future work; not load-bearing for #328 (Qwen3's `model_type=qwen3` is unambiguous).

### DIFF
--- a/pie/src/pie/capabilities.py
+++ b/pie/src/pie/capabilities.py
@@ -60,8 +60,11 @@ class DriverCapabilities:
     max_batch_size: int
 
     # ── Model-architecture facts ───────────────────────────────────────
-    # First architecture string from the HF config (e.g. "LlamaForCausalLM",
-    # "Qwen3ForCausalLM"). Forwarded to Rust as ModelConfig.arch_name.
+    # Pie internal arch name (lowercase: "qwen3", "llama3", …) — the
+    # dispatch key the Rust runtime expects. Drivers MUST resolve this
+    # via `pie_driver.model.resolve()` from the HF config; raw HF class
+    # names like "Qwen3ForCausalLM" are NOT accepted on the Rust side.
+    # See pie_driver/model/__init__.py and pie-project/pie#328.
     arch_name: str
 
     # Vocabulary size and max context length. Currently Python-only —
--- a/pie/src/pie_driver/hf_utils.py
+++ b/pie/src/pie_driver/hf_utils.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 from huggingface_hub.constants import HF_HUB_CACHE
 
-from .model import HF_TO_PIE_ARCH
+from .model import HF_TO_PIE_ARCH, resolve, UnknownArchitectureError
 
 
 
@@ -118,10 +118,8 @@ def check_pie_compatibility(config: dict | None) -> tuple[bool, str]:
     """
     if config is None:
         return False, "no config"
-
-    model_type = config.get("model_type", "")
-    if model_type in HF_TO_PIE_ARCH:
-        return True, HF_TO_PIE_ARCH[model_type]
-
-    return False, f"unsupported type: {model_type}"
+    try:
+        return True, resolve(config)
+    except UnknownArchitectureError:
+        return False, f"unsupported type: {config.get('model_type', '')}"
 
--- a/pie/src/pie_driver/loader.py
+++ b/pie/src/pie_driver/loader.py
@@ -62,22 +62,20 @@ class ModelLoader:
         # Load config from HuggingFace config.json
         hf_config = hf_utils.load_hf_config(self.snapshot_dir)
 
-        # Derive architecture from HF model_type
-        hf_model_type = hf_config.get("model_type", "")
-        # Handle case where model_type might be mapped (e.g. llama -> llama3)
-        # We need hf_utils.HF_TO_PIE_ARCH to map it.
-        # If it's not in the map, check if it's already a valid PIE type (less likely but possible)
-        arch_type = hf_utils.HF_TO_PIE_ARCH.get(hf_model_type)
-
-        if arch_type is None:
-            # Basic fallback or error
-            if hf_model_type in hf_utils.HF_TO_PIE_ARCH.values():
+        # Derive pie internal arch name from HF identity. Single source of
+        # truth: pie_driver.model.resolve().
+        from .model import resolve as resolve_pie_arch, UnknownArchitectureError, HF_TO_PIE_ARCH
+        try:
+            arch_type = resolve_pie_arch(hf_config)
+        except UnknownArchitectureError as e:
+            # Soft fallback for callers that already pass a pie internal name
+            # via `model_type` (legacy behavior — kept until external callers
+            # migrate). Strict path is `resolve()`.
+            hf_model_type = hf_config.get("model_type", "")
+            if hf_model_type in HF_TO_PIE_ARCH.values():
                 arch_type = hf_model_type
             else:
-                raise ValueError(
-                    f"Unsupported HuggingFace model_type: '{hf_model_type}'. "
-                    f"Supported types: {list(hf_utils.HF_TO_PIE_ARCH.keys())}"
-                )
+                raise ValueError(str(e)) from e
 
         # Inject PIE type into config for runtime to use
         hf_config["type"] = arch_type
--- a/pie/src/pie_driver/model/__init__.py
+++ b/pie/src/pie_driver/model/__init__.py
@@ -58,69 +58,124 @@ class Buffer(ABC):
 # =============================================================================
 
 
-REGISTRY: dict[str, ModuleType] = {}
-"""Architecture name → module with (ModelConfig, ForwardPass, create_kv_cache, create_adapter_cache)."""
+REGISTRY: dict[str, str | ModuleType] = {}
+"""Architecture name → module path string (lazy) or loaded module (cached after first get_module)."""
 
 HF_TO_PIE_ARCH: dict[str, str] = {}
 """HuggingFace model_type → PIE architecture name."""
 
 
+class UnknownArchitectureError(ValueError):
+    """Raised when an HF config can't be mapped to a pie internal arch."""
+
+
 def register(
     name: str,
-    module: ModuleType,
+    module_path: str,
     *,
     aliases: tuple[str, ...] = (),
     hf_model_types: tuple[str, ...] = (),
 ):
-    """Register a model architecture.
+    """Register a model architecture (lazy — module loads on first ``get_module``).
 
     Args:
-        name: Primary architecture name (e.g., "llama3")
-        module: Module containing ModelConfig, ForwardPass, create_kv_cache, create_adapter_cache
-        aliases: Additional names that map to the same module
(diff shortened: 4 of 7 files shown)


shsym:
Opened #334 as the discussion thread for the broader "where should HF→pie arch resolution live?" question and the latent silent-fallback follow-ups. This PR focuses on the minimum needed to close #328; the structural questions are noted there for separate discussion.

Issue #327, opened 2026-04-30
[vllm extras] vllm is unpinned in pyproject.toml; PyPI HEAD wheel breaks any CUDA-12 base


shsym:
## `pie-server[vllm]` extra is unpinned, resolves to a CUDA-13-only wheel

- Branch: `features/deva-integration`

`pie/pyproject.toml:83-85` has a bare `vllm` with no version (and the commented sibling-checkout `path = "../../vllm"` at line 141 is the intended pin pattern, but doesn't apply for forks/CI without a sibling). On a clean checkout `uv sync --extra vllm` resolves to **vllm 0.20.0** from PyPI, whose default wheel is built for CUDA 13.0 / libcudart.so.13 (per [vllm's docs](https://docs.vllm.ai/en/stable/getting_started/installation/gpu/) — the +cu129 and +cpu variants ship as separate release assets).

On any CUDA 12.x base:

```
$ uv sync --extra cu128 --extra vllm
$ python -c "import vllm._C"
ImportError: libcudart.so.13: cannot open shared object file: No such file or directory
```

Host driver caps matter too — 575.x maxes at CUDA 12.9 and 570.x at 12.8, so anything cu130 also fails at runtime unless the host driver is 580.x+.

Wanted to check: which vllm version + CUDA generation does `features/deva-integration` actually test against? Happy to put up a quick PR once we know.

One option that worked in our fork (cu129 stack on a 12.9 host):

```toml
vllm = [
    "vllm @ https://github.com/vllm-project/vllm/releases/download/v0.20.0/vllm-0.20.0+cu129-cp38-abi3-manylinux_2_31_x86_64.whl ; sys_platform == 'linux' and platform_machine == 'x86_64'",
]
```

paired with a `cu129` extras section parallel to the existing `cu128` and a `pytorch-cu129` index. Other shapes (PyPI `vllm==0.X.Y`, or a documented sibling-checkout ref) work too — whatever fits the team's workflow.

Hit this while integrating into a clean Docker image; #326 has a related rand_mv lazy-compile + gencode auto-detect fix in the same blast radius.

shsym:
@ingim, could you check this and let us know which versions we should target?

ingim:
What's your take on this? I think the issue is with the wheel selection rather than version 0.20.0 itself. Ideally, we'd align pie's CUDA wheel selection with vllm's.
Since we're using uv, we could map pie's flags to specific vllm wheels, similar to what we did for the flashinfer cache.

[...]

PR #331, opened 2026-04-30
chore(stack): cu129 + headless-vllm + first-call warmup fixes for RTX 4090 / sm_89

shsym:
## What

Three commits that together let `pie` actually build and run smoothly on cu129 with the headless-vllm wheel pin. Carried in `shsym:features/vllm-opt`; rebased onto current upstream HEAD.

```
b6a10355  build(deps): add cu129 extras and adopt headless vllm
fdc22bb7  fix(server):  pre-warm torch.cuda.device_count in main thread
249d045e  fix(runtime): bump fire_batch timeout 30s -> 300s for first-call warmup
```

Diff stat: ~+51 / -8 across 3 files (`pie/pyproject.toml`, `pie/src/pie/server.py`, `runtime/src/device.rs`).

> Originally I bundled the rand_mv compute-capability auto-detect commit too, then noticed it's already in flight as #326 — dropped from this PR to avoid duplication. Treat #326 as a soft prerequisite: without it, this stack still builds and starts, but rand_mv won't bind on any GPU outside the originally-baked target arch.

### `b6a10355` — pyproject: cu129 extras + headless vllm
Adopts the headless-vllm direction discussed in #327. `pie` no longer pins `vllm` directly; the extras section adds cu129 wheels for torch / torchao / flashinfer routed via `--extra-index-url cu129` so transitive `torchvision`/`torchaudio` pulls don't drift to cu130. The container recipe layers vllm on top explicitly (matches the deva-trial Dockerfile pattern).

### `fdc22bb7` — server: pre-warm torch.cuda.device_count
First-call into `torch.cuda` from a non-main thread can stall on CUDA initialization (lazy init crosses the GIL boundary in awkward ways). Pre-warming `torch.cuda.device_count()` in the main thread inside `Server.__aenter__` removes the stall. Surfaced as a multi-second latency spike on the *first* request that touched CUDA — invisible after the first call.

### `249d045e` — runtime: fire_batch timeout 30s → 300s for first-call warmup
The first `fire_batch` RPC bundles vllm engine init + Qwen3-0.6B weight load + flashinfer workspace alloc + torch.compile. On a cold cache that's 60-120s typical, more under contention. The 30s cap was a debugging leftover; 300s is generous enough to not false-trip while still catching genuine wedges.

## Why one PR

The three touch different files but address one composite goal: make pie buildable + non-flaky on cu129. Splitting into three would each individually look unmotivated ("why this constant?", "why this pre-warm?") because the failure mode for each is only visible when running the full stack on the target hardware. Happy to break up if maintainers prefer.

## Validation

These commits have been carried in `shsym:features/vllm-opt` and exercised end-to-end on ECL RTX 4090s for the past several weeks. The most recent e2e was the #328 reproducer ([#330](https://github.com/pie-project/pie/pull/330)) — that test ran on a build of `features/vllm-opt + #330's fix` and the engine came up cleanly.

## Relationship to other PRs / issues

- **#326** — rand_mv compute-capability auto-detect. Soft prereq for actually loading on sm_89; not duplicated here.
- **#327** — headless-vllm direction discussion; this PR ships the concrete pyproject change.
- **#330** — the #328 reasoning bug fix, independent of this stack (touches different files, no overlap).
- No conflicts with current `features/deva-integration` HEAD (`74787e61`). `cargo check` clean.

## Out of scope

- The actual vllm install (still container-side; matches the headless-vllm contract — pie ships pyproject without `vllm` as a direct dep).
- Any of the multistep / async-overlap / KV admission work from `shsym:features/vllm-opt`'s history that DIDN'T land here — those are separate experimental branches.

### DIFF
--- a/pie/pyproject.toml
+++ b/pie/pyproject.toml
@@ -66,6 +66,11 @@ cu128 = [
     "flashinfer-jit-cache==0.6.8.post1",
 ]
 
+cu129 = [
+    "torch>=2.11.0",
+    "torchao>=0.14.1",
+]
+
 # --- Metal (macOS) ---
 metal = [
     "torch>=2.7.1",
@@ -77,12 +82,13 @@ metal = [
 ]
 
 # --- vLLM backend (CUDA / ROCm / XPU / TPU) ---
-# Pinned tight to the version we test against; bumped weekly via automation.
-# Sourced from the sibling `vllm/` checkout so we track upstream tip, not
-# the stale `vllm` PyPI metadata (the cu128 mirror lags far behind upstream).
-vllm = [
-    "vllm",
-]
+# Headless: pie does NOT pin vllm. Users install vllm themselves per
+# https://docs.vllm.ai/en/stable/getting_started/installation/gpu/, picking
+# the wheel that matches their CUDA generation (e.g. +cu129 release asset
+# for a CUDA 12.9 host, the default PyPI wheel for cu130 hosts). pie's
+# pie_driver_vllm raises a clear error at import time if vllm is missing.
+# Supported range: vllm>=0.20,<0.21 as of this writing.
+vllm = []
 
 # --- SGLang backend (CUDA / ROCm / XPU / NPU) ---
 # Sourced from the sibling `sglang/` checkout so we track upstream tip.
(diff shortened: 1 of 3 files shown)


shsym:
Closing in favor of two focused PRs split by concern: cu129+headless-vllm decoupling (per #327 Option 1) vs first-call warmup hygiene fixes. Links to follow.

PR #362, opened 2026-05-16
driver/portable: load MoE experts from stacked GGUF tensors

shsym:
## TL;DR

* Make MoE GGUFs load: probe for stacked expert slabs, route through `declare_stacked_experts_from_3d_`.
* Fix a pre-existing GGUF blocker: `declare_` was misreading native-F16 GGUF tensors as F32-source.
* Two changes, one file (`driver/portable/src/model.cpp`).

## Why

GGUF stores MoE experts as three stacked 3D tensors (gate / up / down), not the per-expert HF safetensors layout that `load_moe_layer_` assumed. Every MoE GGUF — qwen3_moe, mixtral, gpt-oss — hit `missing tensor` at build time. Separately, `declare_` keyed its cast logic on `StDtype`, which is a placeholder F32 for every GGUF tensor, so native-F16 tensors like `token_embd.weight` crashed at a 2× size mismatch.

## What

`load_moe_layer_` probes for `mlp.experts.gate_proj.weight` (the stacked name the GGUF mapper emits uniformly across families). When present, gate/up/down are loaded via the existing `declare_stacked_experts_from_3d_` helper (same helper qwen3.5-moe's fused safetensors path already uses). Per-expert safetensors path is unchanged and reached as a fallthrough.

`declare_` skips upcast/downcast when `ggml_type_override >= 0` — same `(t.ggml_type_override < 0 ? safetensors : gguf)` convention already used in `new_tensor_from_st`.

## Smoke test

Unsloth's `Qwen3-30B-A3B-Q4_K_M.gguf` on Apple M4 Pro / Metal:

```
[pie-driver-portable] loaded qwen3_moe (48 layers, hidden=2048, heads=32/4 kv,
                                          buf=18117.9 MiB, backend=Apple M4 Pro)
[pie-driver-portable] forward engine ready (total_pages=1024, kv_buf=3072 MiB)
[pie-driver-portable] serving on shmem /pie_shmem_g0
```

Pairs with #361 (the user-facing `.gguf` `hf_repo` entry point) but is independent.

### DIFF
--- a/driver/portable/src/model.cpp
+++ b/driver/portable/src/model.cpp
@@ -494,13 +494,21 @@ ggml_tensor* Model::declare_(const std::string& hf_name) {
     // tensor we hand back to the graph builder is typed F32, so the
     // implicit `ggml_cast` in `norm_scale` becomes a no-op. The actual
     // bf16 → f32 byte conversion happens in load_into_backend_().
-    const bool upcast = t.dtype == StDtype::BF16
+    //
+    // GGUF tensors carry their ggml type in `ggml_type_override` and
+    // leave `dtype` as a placeholder F32 (see gguf_archive). The cast
+    // logic below is keyed off `dtype` and would mistrigger on every
+    // GGUF tensor, so we only apply it on the safetensors path.
+    const bool is_safetensors = t.ggml_type_override < 0;
+    const bool upcast = is_safetensors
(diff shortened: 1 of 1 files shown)


shsym:
Preparing review...

shsym:
Preparing review...

PR #370, opened 2026-05-19
Show explicit unsupported-driver error on Windows

### LINKED ISSUE #366: Dev driver setup fails on Windows due to incorrect Python interpreter path
## TL;DR

The documentation suggests installing the dev driver with:

```bash
pie driver dev install ~/.pie/venvs/dev --run
```

On Windows, this command creates the virtual environment successfully, but the install step then tries to use a Unix-style interpreter path:

```text
~/.pie/venvs/dev\bin\python
```

Windows virtual environments use `Scripts\python.exe`, so the install step fails.

## Repro

On Windows PowerShell, run the command from the documentation:

```powershell
pie driver dev install ~/.pie/venvs/dev --run
```

## Actual behavior

The venv is created successfully:

```text
Activate with: ~\.pie\venvs\dev\Scripts\activate
```

But the next step uses:

```text
uv pip install --python ~/.pie/venvs/dev\bin\python pie-driver-dev[cu128]
```

and fails with:

```text
error: No virtual environment or system Python installation found for path `~/.pie/venvs/dev\bin\python`
```

## Expected behavior

On Windows, Pie should either:

- use the Windows venv interpreter path:

```text
~/.pie/venvs/dev\Scripts\python.exe
```

## Investigation

The interpreter path appears to be constructed in:

```text
server/src/cli/driver_cmd.rs
```

with:

```rust
venv.join("bin").join("python")
```

This works for Unix-like virtual environments, but not on Windows.

## Environment

- OS: Windows
- Shell: PowerShell
- Pie version: 0.3.0
- Command: `pie driver dev install ~/.pie/venvs/dev --run`

LIU-ZHIYUAN-source:
This PR adds an explicit Windows guard for subprocess drivers (dev, vllm, sglang).

Previously, running `pie driver dev install` on Windows would print a Unix-style interpreter path (`<venv>/bin/python`), which is misleading since Windows uses `Scripts/python.exe`.

Since these drivers are currently unsupported on Windows, this change fails early with a clear error message and suggests using the portable driver on Windows or running these drivers in WSL/Linux instead.

Tested on Windows:
- cargo run -p pie-server --no-default-features -- driver dev install

Fixes #366 

## Summary by CodeRabbit

* **Bug Fixes**
  * Driver install now immediately reports incompatibility on Windows for the dev, vllm, and sglang drivers and directs users to use the portable driver or run those drivers in WSL/Linux.

* **Documentation**
  * Added a Windows compatibility note in the installation guide under "Optional: Python-backed drivers" with guidance on supported alternatives.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pie-project/pie/pull/370?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/cli/driver_cmd.rs
+++ b/server/src/cli/driver_cmd.rs
@@ -217,6 +217,14 @@ fn run_subprocess(flavor: SubprocessFlavor, action: PerDriverCmd) -> Result<()> 
 }
 
 fn install(flavor: SubprocessFlavor, path: Option<PathBuf>, run: bool) -> Result<()> {
+    #[cfg(windows)]
+    if matches!(flavor, SubprocessFlavor::Dev | SubprocessFlavor::Vllm | SubprocessFlavor::Sglang) {
+        bail!(
+            "{} driver is not supported on Windows. Please use the portable driver on Windows, or use WSL/Linux for dev, vllm, or sglang drivers.",
+            flavor.as_str()
+        );
+    }
+    
     let venv = path.unwrap_or_else(|| crate::paths::pie_home().join("venvs").join(flavor.as_str()));
     let python_bin = venv.join("bin").join("python");
 
--- a/website/docs/guide/install.mdx
+++ b/website/docs/guide/install.mdx
@@ -70,6 +70,9 @@ This provides the standalone `bakery` command and the Python package used by
 
 The `pie` binary runs the embedded `cuda_native` and `portable` drivers without Python. Python is required only for the subprocess drivers: `dev`, `vllm`, and `sglang`. Install one with:
 
+> Note: The `dev`, `vllm`, and `sglang` drivers are currently not supported on Windows.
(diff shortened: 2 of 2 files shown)


LIU-ZHIYUAN-source:
Hi @ingim , quick question — the docs show install commands for dev/vllm/sglang, but don’t mention Windows support.

Would it make sense to clarify in the docs that these drivers are currently unsupported on Windows, or just leave it as is?

ingim:
Updating the doc makes sense, would you be able to include the doc change in the pr?

LIU-ZHIYUAN-source:
> Updating the doc makes sense, would you be able to include the doc change in the pr?

Thanks! I've added a note in the docs clarifying that dev/vllm/sglang drivers are currently unsupported on Windows.

ingim:
Great, thank you!

PR #386, opened 2026-06-05
feat(runtime): max-output-tokens follows scheduler default_token_limit

shsym:
Sibling PRs for this ticket:
- pie-project/pie#385 — https://github.com/pie-project/pie/pull/385
- shsym/RatioThink#50 — https://github.com/shsym/RatioThink/pull/50

## What
`runtime::max-output-tokens()` now returns the configured `[model.scheduler].default_token_limit` when set, falling back to the raw KV capacity otherwise (minimum across registered models). The WIT signature is unchanged — only the value source changes.

## Why
Follow-up to #385. `default_token_limit` is the scheduler's total-token compute cap (prompt + output, enforced per forward pass). Surfacing it through `max-output-tokens` lets a host derive the inferlet's per-request ceiling from a memory-aware compute budget computed at launch — instead of the fixed KV-pool capacity. Inferlets (chat-apc) already read this import for their clean-400 boundary, so no inferlet change is needed.

## Changes
- `model::register` stores the per-model `default_token_limit: Option<u32>`.
- `bootstrap` threads `cfg.scheduler.default_token_limit` through (saturated to u32).
- `min_output_token_ceiling()` = `min` over models of `default_token_limit ?? kv_capacity_tokens`; `api/runtime::max_output_tokens` returns it.

## Testing
- `cargo check -p pie` + `cargo test -p pie --lib` (392 passed, 0 failed).
- Exercised end-to-end by the RatioThink consumer: a dummy-driver engine with `default_token_limit` set returns it as the `max_tokens` ceiling in the over-limit 400 message.

### DIFF
--- a/runtime/src/api/runtime.rs
+++ b/runtime/src/api/runtime.rs
@@ -24,6 +24,6 @@ impl pie::core::runtime::Host for InstanceState {
     }
 
     async fn max_output_tokens(&mut self) -> Result<u32> {
-        Ok(model::min_kv_capacity_tokens())
+        Ok(model::min_output_token_ceiling())
     }
 }
--- a/runtime/src/bootstrap.rs
+++ b/runtime/src/bootstrap.rs
@@ -219,11 +219,19 @@ async fn bootstrap_inner(config: Config, listener: Option<TcpListener>) -> Resul
             .min()
             .unwrap_or(0)
             .min(u32::MAX as u64) as u32;
+        // Scheduler compute cap (total tokens / process). When set it is
+        // the per-request output ceiling the inferlet follows; saturate
+        // into u32 for the WIT surface.
+        let default_token_limit = cfg
+            .scheduler
+            .default_token_limit
+            .map(|n| n.min(u32::MAX as usize) as u32);
         model::register(
             cfg.name.clone(),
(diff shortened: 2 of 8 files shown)


ingim:
`max-output-tokens` and `default_token_limit` serve different purposes. default_token_limit is just a scheduler hint how many tokens an inferlet can generate without being interrupted. Ie., inferlets can generate more tokens than the value if the server is uncontended. `max-output-tokens` is a hard physical limit.

Maybe I could improve the variable names to avoid this confusion

shsym:
Clarification after the RatioThink PR discussion:

What is currently missing in Pie is an inferlet-facing hard output ceiling that reflects the engine's launch-time memory budget. Inferlets should not read scheduler policy directly; they should read `runtime::max-output-tokens`.

This PR adds that bridge: Pie keeps `default_token_limit` as scheduler/process-budget configuration, and exposes the effective output ceiling through `runtime::max-output-tokens`, capped by physical KV capacity. The RatioThink/chat-apc side then consumes only that runtime API.

shsym:
I will merge this as it is toward the non-main branch, and this is required for the mac app's http daemon inferlets.

PR #400, opened 2026-06-08
MCTS, Modular-Cache, Hierarchical-Attention Inferlet Examples

agmankaruse:
### Summary

This PR adds three new Pie example inferlet families: modular-cache, hierarchical-attention and MCTS reasoning.

The goal is to show a few more of the programmable-serving patterns described in the Pie paper, using small examples that are easy to build, run, and inspect. Each example is implemented across Rust, Python, and JavaScript so the same idea can be compared across SDKs.

The MCTS example demonstrates an iterative reasoning loop. It uses selection, expansion, rollout, evaluation and backpropagation to build a small reasoning tree before synthesizing a final answer.

The modular-cache example demonstrates application-controlled KV cache reuse. It breaks a prompt into dependency-ordered modules, saves snapshots after stable module prefixes and reuses the longest matching prefix when later launches share the same persistent `pie serve` engine.

The hierarchical-attention example demonstrates structured attention control. It chunks a prompt, keeps global instructions and chunk summaries visible, selects the most relevant chunk by lexical overlap and applies a BRLE attention mask during manual forward passes.

### Tests

I added host-side pure-logic tests for the parts that do not need a running Pie engine, including modular-cache ordering/keying, hierarchical-attention chunking and masks, and MCTS scoring/tree logic.

I also added E2E-style runners that launch the inferlets against the dummy driver and assert on structured control-flow output. These tests check that the examples run correctly and emit the expected traces, rather than trying to judge model quality.

## Release Notes

* **New Features**
  * Added hierarchical-attention inferlets for efficient prompt processing using hierarchical attention masks.
  * Added Monte Carlo Tree Search (MCTS) inferlets for reasoning-based generation tasks.
  * Added modular-cache inferlets for KV cache snapshots and reuse across inference runs.
  * All new inferlets available in JavaScript, Python, and Rust implementations.

### DIFF
--- a/inferlets/hierarchical-attention-js/Pie.toml
+++ b/inferlets/hierarchical-attention-js/Pie.toml
@@ -1,0 +1,21 @@
+[package]
+name = "hierarchical-attention-js"
+version = "0.1.0"
+description = "TypeScript hierarchical attention mask demo over chunk summaries plus selected local chunk"
+authors = ["Abraam Mankaruse"]
+
+[runtime]
+core = "^0.2.0"
+mcp = "^0.2.0"
+
+[parameters]
+prompt = {type = "string", optional = true, description = "User prompt"}
+max_tokens = {type = "int", optional = true, description = "Maximum number of output tokens"}
+
+chunk_size_words = {type = "int", optional = true, description = "Approximate words per chunk"}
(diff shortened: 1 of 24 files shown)


ingim:
Great work, @agmankaruse!

PR #406, opened 2026-06-09
Add AsyncLM example inferlet (asynchronous LLM function calling)

evanc7007:
## Summary

  Adds `inferlets/async-lm`, a Pie inferlet implementing **AsyncLM** (Gim, Lee,
  Zhong — *Asynchronous LLM Function Calling*, [arXiv:2412.07017](https://arxiv.org/abs/2412.07017)).
  The model dispatches tool calls **without blocking** and keeps generating, so
  function-call latency overlaps token decoding instead of stalling on each call.

  Model ↔ runtime communicate over the paper's CML (Context Markup Language)
  token protocol:

  - `[CALL] <id> [HEAD] <code> [END]` — dispatch a call (non-blocking)
  - `[TRAP][END]` — pause generation until every dispatched call has returned
  - `[INTR] <id> [HEAD] <value> [END]` — runtime-injected result frame (read-only
    to the model)

  The inferlet drives the whole loop: an FSM parser scans the token stream for CML
  delimiters, each `[CALL]` fires as a real HTTP request on the `wstd` reactor, and
  each `[INTR]` result is injected back into the live `Generator`.

  ## Implementation

  - **CML FSM parser** — tracks Normal/Call/Trap/Intr states, handles BPE-merged
    delimiters (`][`, `]\n`, a leading `[` fused into the previous token), and
    bypasses CML matching inside `<think>…</think>` so protocol text the model
    quotes while reasoning isn't acted on.
  - **Five live tools** — weather, stock price, currency, time, Wikipedia, each a
    real web API; calls run concurrently with decode.
  - **`[INTR]` suppression** — a precomputed BRLE constraint mask bans the token
    ids unique to `[INTR]`, so the model can't forge a result frame; shared
    brackets stay sampleable.
  - **Trap-time scheduling** — a Keep/Recompute/Swap classifier backed by
    `[CALL]`-site checkpoints, a critical-section tracker so results inject only at
    safe parser boundaries, and optional mid-stream injection (deliver a result
    *before* the model reaches `[TRAP]`).
  - **Recompute** restores a `Context` snapshot and replays the completed frames.
    **Swap** is classified/logged but executes as Keep (see SDK limitations).

  Pie features exercised: `Context::{snapshot, open, delete}` for
  checkpoint/restore, token-level `Generator` driving
  (`next`/`execute`/`accept`/`clear_sampler`), `Constrain` masks, and `wstd` async
  HTTP.

  ## Diagnostic flags (switchable)

  Per review feedback, the instrumentation is kept but **off by default and
  toggled per-run** rather than hard-wired:

  | Flag | Default | Effect |
  |---|---|---|
  | `enable_trace` | false | machine-readable `[TRACE]` event stream for offline latency analysis (feeds the Gantt below) |
  | `enable_token_trace` | false | verbose per-token `[dbg]` decode/injection diagnostics |
  | `disable_checkpointing` | true | single unbroken generator session, always-Keep (max overlap); set false for paper-faithful checkpoint/restore + Recompute |
  | `emulate_unknown_tools` | false | treat non-live tools as BFCL latency-emulated calls instead of erroring |

  ## Results

  Measured on `no_think` (see limitations re: thinking mode). Latency is
  end-to-end task completion, async vs. synchronous and sync-parallel baselines,
  on a prompt set with genuine overlap opportunity (multi-step / dependent calls).

  **Positive — async overlaps function latency with decode and wins on
  multi-step / dependent-call workloads**, matching the paper's mechanism:

Async Benchmarks

  Per-call timeline (`enable_trace` → Gantt) showing calls completing *during*
  decode rather than at the trap:

p0_r03

  **Negative / where async doesn't help (and why):**

  - **Single-turn parallel lists → ~no win.** When the model dispatches every
    independent call in one batch and immediately traps, there's no prose window
    to overlap and (with short tool latency) nothing to hide. The win needs
    dependent multi-round structure or genuinely long tool latency.
  - **Reliable CML emission requires a sizable few-shot system prompt.** Without
    the default `system` prompt's rules + examples, models drift — mixing prose
    into `[CALL]` bodies, inventing calls for knowledge-only questions, or echoing
    `[INTR]` frames into output. *Cause: the CML control tokens are
    out-of-distribution and the model has no native async-dispatch notion, so the
    protocol has to be fully specified in-context. Smaller/quantized models need
    more guidance — e.g. mid-stream `[INTR]` echo is reliable on Qwen3-30B-A3B
    (Q4_K_M) but rare on Qwen3-235B-A22B (Q4_K_M).*
  - **Thinking mode is high-variance.** With Qwen3 thinking on, the model variably
    stalls inside `<think>` planning the CML before emitting any call, sometimes
    exhausting the token budget; `no_think` is recommended. *Cause: thinking
    interacts poorly with the strict output protocol — the model deliberates about
    the format instead of executing it, and deliberation length is unbounded.*
    Async still functions with thinking; the latency data is just noisy.

  ## SDK limitations surfaced

  - **No RAM-tier KV swap primitive** → Swap collapses to Keep.
  - **`Context::take()`** on active/on-GPU snapshots currently needs extra GPU
    pages; Recompute uses `Context::open()` + best-effort delete as a workaround.

  ## Testing

  Live-tested against `pie serve` (Qwen3-8B, plus remote Qwen3-30B-A3B and
  Qwen3-235B-A22B): parallel calls, dependent chains, checkpoint save/restore,
  `[TRAP]` drain + `[INTR]` injection, Keep/Recompute/Swap paths, mid-stream
  injection, and all four flags in both states.

  ## Files

  `inferlets/async-lm/{Cargo.toml, Pie.toml, src/lib.rs}` — standard inferlet
  layout (`[runtime]` + `[parameters]` manifest; `Cargo.lock` / `target`
  gitignored).

## Summary by CodeRabbit

* **New Features**
  * Added an async language-model component that runs tools concurrently over HTTP and injects their results into streaming output without blocking.
  * Exposed runtime generation controls (temperature, top-p, max-tokens, optional step ceiling) and cost-model tuning for interrupt/trap behavior.
  * Added checkpoint/restart support, mid-stream injection, token-level tracing/diagnostics, interrupt suppression, and an option to emulate unknown tools.

### DIFF
--- a/inferlets/async-lm/Cargo.toml
+++ b/inferlets/async-lm/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "async-lm"
+version = "0.1.0"
+edition = "2024"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+inferlet = { path = "../../sdk/rust/inferlet" }
+serde = { version = "1.0", features = ["derive"] }
+
+[profile.release]
+lto = true
--- a/inferlets/async-lm/Pie.toml
+++ b/inferlets/async-lm/Pie.toml
@@ -1,0 +1,25 @@
+[package]
+name = "async-lm"
+version = "0.1.0"
+description = "AsyncLM: asynchronous LLM function calling — non-blocking tool calls that overlap decoding (arXiv:2412.07017)"
+authors = ["Pie Team"]
+repository = "https://github.com/pie-project/pie"
+
+[runtime]
+core = "^0.2.0"
+mcp = "^0.2.0"
+
+[parameters]
+prompt = {type = "string", description = "The user message to answer"}
+system = {type = "string", optional = true, description = "System prompt to set assistant behavior (default: AsyncLM/CML protocol instructions)"}
+max_tokens = {type = "int", optional = true, description = "Maximum number of tokens to generate (default: 2048)"}
+temperature = {type = "float", optional = true, description = "Sampling temperature (default: 0.6)"}
+top_p = {type = "float", optional = true, description = "Top-p nucleus sampling threshold (default: 0.95)"}
+alpha_recompute_ns_per_tok2 = {type = "float", optional = true, description = "Recompute cost-model alpha in ns/token^2 (default: 800)"}
+beta_swap_ns_per_tok = {type = "float", optional = true, description = "Swap cost-model beta in ns/token (default: 400000)"}
+enable_midstream = {type = "bool", optional = true, description = "Enable safe-boundary mid-stream [INTR] injection — deliver a result before the model reaches [TRAP]; default false"}
+enable_trace = {type = "bool", optional = true, description = "Emit machine-readable [TRACE] timeline events for offline latency analysis; default false"}
+enable_token_trace = {type = "bool", optional = true, description = "Emit verbose per-token [dbg] decode/injection diagnostics; default false"}
+disable_checkpointing = {type = "bool", optional = true, description = "Skip per-call checkpointing and force trap strategy Keep (one unbroken generator session); default true (set false for paper-faithful checkpoint/restore + Recompute)"}
+emulate_unknown_tools = {type = "bool", optional = true, description = "Treat unknown tools as BFCL latency-emulated calls instead of erroring; default false"}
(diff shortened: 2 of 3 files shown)


evanc7007:
I can modularize the code into different sections if needed.

ingim:
Great work, @evanc7007 !

PR #410, opened 2026-06-11
fix: paged-attn decode planner padded_batch_size/tile-count mismatch

evanc7007:
DecodeWorkEstimator forces split_kv=false (driver/portable: always; driver/cuda: for head_dims outside {64,128,256,512}) without adjusting the work estimator's max_num_pages_per_batch (KV-chunk size). Under split_kv=false, DecodePlan sets padded_batch_size = batch_size, but DecodeSplitKVIndptr still tiles each request's KV pages using the unadjusted (split-kv-sized) chunk size, producing more (request, tile) work items than padded_batch_size whenever a request's page count exceeds that chunk size. The resulting std::copy overflows the request_indices/kv_tile_indices/o_indptr buffers, causing multiple grid blocks to map onto the same request and clobbering other requests' output rows.

This was the root cause of severe output corruption (duplicated tokens, cross-request marker bleed, reasoning-channel leakage) whenever 2+ concurrent requests with distinct prompts/KV-page-counts were batched for paged decode.

Fix: bump max_num_pages_per_batch to the largest per-request KV page count whenever split_kv is forced false, guaranteeing exactly one tile per request so request_indices_vec.size() == batch_size == padded_batch_size. kv_chunk_size_ptr is otherwise unused when partition_kv == false, so this only affects planner bookkeeping.

Verified on driver/portable with Qwen3-30B-A3B: 0 corrupted outputs across conc=2-4, ~25-6000 chars/request (was ~50-100% corrupted before the fix). The driver/cuda fix is the same change applied by analogy (affects head_dims outside {64,128,256,512}, e.g. Phi-3-mini); not yet build/test-verified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected attention decode work-estimation to compute per-request resource needs and adjust tiling accordingly, preventing buffer over-allocation and crashes in certain hardware/configuration scenarios.

### DIFF
--- a/driver/cuda/src/ops/attention_flashinfer.cu
+++ b/driver/cuda/src/ops/attention_flashinfer.cu
@@ -1,5 +1,6 @@
 #include "ops/attention_flashinfer.hpp"
 
+#include <algorithm>
 #include <type_traits>
 
 #include <cmath>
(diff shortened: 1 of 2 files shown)


ingim:
Great find, @evanc7007 !

PR #413, opened 2026-06-12
bring #409, #410 into v1-base-shmem

evanc7007:
Fixes the engine concurrency issues found in chat-apc. Brings the fixes that were already applied in main.


evanc7007:
@shsym, this is just bringing the concurrency fixes into this branch for chat-apc.

PR #424, opened 2026-06-16
feat(portable): surface cause of any fatal driver-thread signal (#691)

shsym:
Sibling PRs for this ticket:
- shsym/RatioThink#268 — https://github.com/shsym/RatioThink/pull/268

## What

Surface the cause of any uncatchable native fault on the embedded driver thread (SIGBUS/SIGSEGV/SIGILL/SIGFPE/SIGABRT), not just the truncated-GGUF SIGBUS the #688 bounds guard covers. Previously such a fault killed the whole host process with no surfaced cause: the C++ run wrappers catch only `std::exception`, host-side exit detection sees only a clean nonzero rc, and the driver is launched with `install_signal_handlers=0`.

## How

- New header-only `driver/portable/src/fatal_backstop.hpp`, installed at the top of `run_impl` before model load (so a load-time fault is caught too). It writes an async-signal-safe `[pie-driver-portable] fatal: caught <SIG> (model=...)` to stderr, then chains to the previously-installed disposition so the process still terminates with the correct signal status. No new FFI surface — surfaces via stderr, leaving `run_inproc`/`InProcVTable` untouched.
- wasmtime coexistence via a thread-local "driver thread" gate: wasmtime executes wasm only on host worker threads, never on the driver thread, so a fault on the driver thread is always a real driver fault (emit + chain to die). On any other thread we emit nothing and chain straight through, so wasm guard-page recovery is byte-for-byte preserved (the saved prev handler's `ctx` is the one the kernel `sigreturn`s into). macOS uses Mach exception ports, so the sigaction handler never even sees an expected wasm trap.

## Tests

`driver/portable/tests/run-fatal-backstop-test.sh` (hermetic, no ggml/Metal build needed): 9/9 covering emit / gate / chain / exit-status across all five signals. Mutation-proven — breaking the gate (always-emit) or the chain (no prev invocation) fails the suite. Also verified in the real Metal build: engine boots + serves (real-model + HTTP e2e) with the backstop installed and no spurious diagnostics.

## Scope

Portable flavor only. cuda can adopt the same header once the Linux/wasmtime guard-page recovery is verified on a Linux host (cannot be exercised on macOS). dummy (Rust, memory-safe) cannot native-fault.

### DIFF
--- a/driver/portable/src/entry.cpp
+++ b/driver/portable/src/entry.cpp
@@ -33,6 +33,7 @@
 #include "host_swap_pool.hpp"
 #include "config.hpp"
 #include "executor/executor.hpp"
+#include "fatal_backstop.hpp"
 #include "gguf_tokenizer.hpp"
 #include "hf_config.hpp"
 #include <pie_bridge/inproc_server.hpp>
@@ -267,6 +268,15 @@ int run_impl(int argc,
     if (!cfg.runtime.verbose) {
         ggml_log_set(quiet_ggml_log, nullptr);
     }
+
+    // Install the fatal-signal backstop before any model work so an
+    // uncatchable native fault during load (e.g. a wild read past a
+    // truncated mmap — the #688 SIGBUS) surfaces its cause to stderr
+    // instead of killing the host process silently. Marking this thread
+    // is per-thread; installing the handlers is process-once. See
+    // fatal_backstop.hpp for the wasmtime-coexistence reasoning.
+    pie_driver_backstop::mark_driver_thread();
+    pie_driver_backstop::install("portable", cfg.model.hf_path.c_str());
 
     // Informational logs go to stderr — stdout is reserved for the READY
     // handshake line consumed by the host process.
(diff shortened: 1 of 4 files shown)


shsym:
Review v1 F2 (worker-thread coverage) addressed in 4eaa9691.

ggml uses its default internal pthread pool with no pie-owned init hook, so there's no worker-entry point to mark directly. But the gate only needs to suppress emission for an *expected, recoverable* wasm guard-page trap, and such a trap can reach a POSIX sigaction handler only where wasmtime owns the signal — i.e. Linux. On Apple, wasmtime uses Mach exception ports, so an expected wasm trap is recovered out-of-band and never delivered as a BSD signal; any signal that reaches the handler there is already a genuine fault on any thread.

So the diagnostic gate is now `t_on_driver_thread || !g_wasm_traps_reach_sigaction` (false on Apple, true on Linux):
- macOS: surfaces an inference-time fault on a ggml compute worker or a Metal completion thread.
- Linux: the strict marked-thread gate is retained for wasm safety; worker coverage there needs a ggml worker-entry hook, deferred with the cuda/Linux follow-up.

The chain to the previous disposition is unchanged on every thread. The hermetic test gains an unmarked worker-thread fault case under Mach-port semantics (diagnostic still emitted + correct exit status), mutation-proven that dropping the new clause silences it; the platform flag is overridable so both regimes run on one host.

F1 (SA_ONSTACK + sigaltstack for stack-overflow SIGSEGV) is filed as a separate ticket per the review routing.

PR #425, opened 2026-06-16
fix: ToT KV-pressure crashes — free pages on context drop + break collect_* on starvation

shsym:
Two complementary fixes for the tree-of-thought daemon crash under KV pressure.

## 1. `fix(context): free KV pages on guest context drop (RAII)` — 528e8bdd

The host WIT context `drop` handler only deleted the wasm resource-table entry and deferred **all** KV-page cleanup to whole-instance `DestroyAll`. A long-lived daemon inferlet (one instance serves every request), and ToT forks many transient contexts per request (`breadth` children + a value-scoring fork per node + spent parents). Those dropped forks' pages were never reclaimed → a heavy tree's fork allocations defer in `when_allocated` waiting on pages only the leaked contexts hold → deadlock.

Free the context eagerly on drop (each WIT `Context` maps 1:1 to a `ContextId`). Snapshots unaffected (separate `owner:None` id). Verified before/after on real portable-Metal: unfixed deadlocks on a heavy tree, fixed completes 3/3.

## 2. `fix(generation): break collect_* on forward-pass starvation` — 2714bd22

The **direct** fix. pie's scheduler folds a device failure / per-batch timeout / mid-stream KV eviction into an empty `Output` (`Ok`, not `Err`). The SDK `Generator::collect_tokens`/`collect_text` sugars folded that as "0 tokens accepted" — never advancing `tokens_generated`, never hitting `max_tokens`/stop — so they **spun**, and the next empty-input forward pass hung the driver ("Server disconnected without sending a response", port dead, driver healthy).

The chat-completions loop already guards this in its own hand-rolled loop, but ToT generates every node + value score via `collect_text`, which had no guard. Add the same starvation predicate (a step whose raw host slots carry no `Token` slot) to the `collect_*` sugars — terminate with a clear error instead of spinning. A natural stop is not starvation (the host still samples `Token(stop)`, later truncated out of `Output::tokens`). Unit-tested (4 cases).

## Tests
- `cargo test --lib` (inferlet SDK): 25 passed incl. 4 starvation-predicate tests.
- Downstream: dummy-driver HTTP e2e 202/1-skip; chat-apc native 106; deep-tree deadlock regression 3/3 (real Metal).

### DIFF
--- a/runtime/src/api/context.rs
+++ b/runtime/src/api/context.rs
@@ -233,18 +233,31 @@ impl pie::core::context::HostContext for InstanceState {
     }
 
     async fn drop(&mut self, this: Resource<Context>) -> Result<()> {
-        // Pull ctx + model identifiers BEFORE we release the table
-        // entry — the resource handle is invalid after delete.
+        // Free the context's KV pages when the guest drops the handle — each
+        // WIT Context resource maps 1:1 to a ContextId, so a dropped handle
+        // means the guest is done with that context. A whole-instance
+        // DestroyAll still runs on instance teardown for anything still live,
+        // but deferring ALL cleanup to teardown leaks every intermediate
+        // context for the lifetime of a long-lived daemon inferlet (chat-apc
+        // serves every request on ONE instance). The tree-of-thought search
+        // forks many transient contexts per request — children, per-node
+        // value-scoring forks, pruned beam losers — and relied on this drop
+        // to reclaim them; without eager freeing the KV pool is exhausted and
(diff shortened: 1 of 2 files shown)


ingim:
Fix 1 is good. But I don't understand the "starvation" mechanism for 2. 

How come this can happen? If this is true, then we need to find and fix the root cause.
> True when a generation step came back with **no sampled token**: the
> /// host returned zero `Token` slots for a decode that had the auto-sampler
> /// attached.

ingim:
@shsym  Could you open a PR for fix 1? I think this should land in main

PR #426, opened 2026-06-17
fix(portable): key uniform-sample reshape on slot count, not request count

shsym:
**Problem:** the portable non-greedy uniform top-k fast path reshapes the per-slot probability tensor with `n_req` as the trailing dim. `probs` has one row per sampling *slot*, and speculation samples >1 slot per request (`n_slots > n_req`), so `ggml_reshape` trips `GGML_ASSERT(nelements...)` and the driver dies mid-stream. Greedy (`argmax`) is per-slot and unaffected, so only non-greedy sampling under speculation (e.g. tree-of-thought at temperature > 0) hits it.

**Fix:** key both reshapes on the actual slot count (`probs->ne[1]`). With one slot per request `n_slots == n_req`, so non-speculative decoding is unchanged. Same defect/fix across all arch graph builders: qwen3, qwen3.5, gemma4, phi3-small, phi3.5-moe.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tensor dimension calculations in top-k sampling operations across Gemma4, Phi3.5-MoE, Phi3-Small, Qwen3, and Qwen3.5 models to correctly handle scenarios with multiple samples per request.

* **Documentation**
  * Updated tensor shape specifications for output indices to reflect correct dimensionality.

### DIFF
--- a/driver/portable/src/graph_common.hpp
+++ b/driver/portable/src/graph_common.hpp
@@ -94,7 +94,7 @@ struct GraphInputs {
     ggml_tensor*              tok_input;   // I32 [total_n_tokens]
     ggml_tensor*              pos_input;   // I32 [total_n_tokens]
     ggml_tensor*              kv_idxs;     // I64 [total_n_tokens] (write idxs)
-    ggml_tensor*              out_idx;     // I32 [n_request]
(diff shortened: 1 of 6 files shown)


ingim:
Great catch

PR #427, opened 2026-06-17
fix(runtime,portable): correct Llama3 portable GGUF RoPE + stop on eom

shsym:
**Problem:** Llama3 GGUFs use ggml's *normal* RoPE pairing (adjacent head values), but the shared qwen graph hard-codes `GGML_ROPE_TYPE_NEOX` for every arch. That rotates Llama Q/K with the wrong dimension pairs, so the model attends incoherently while tokenization/template look correct (e.g. Llama 3.1 answers a simple prompt with generic boilerplate). Separately, generation only stopped on `<|eot_id|>`/`<|end_of_text|>`, so Llama 3.1+ tool/ipython turns that end with `<|eom_id|>` ran on.

**Fix:** pick `rope_type` by arch (`Llama3` -> normal, else NEOX) in the portable graph; add `<|eom_id|>` to the Llama3 stop set. Regression: `has_correct_stop_tokens` now expects 3 stop ids.

Note: the fork carry-patch also rewired the Llama3 prompt prologue, but `main` already builds that via the Jinja chat template, so only the RoPE + eom parts apply here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rotational position embedding handling by selecting the correct RoPE variant per model architecture, improving attention behavior across different models.
  * Improved Llama3 instruction generation stopping accuracy by adding `<|eom_id|>` as an additional end marker.
* **Tests**
  * Updated the unit-test tokenizer vocabulary and stop-token count expectations to cover the new `<|eom_id|>` stop token.

### DIFF
--- a/driver/portable/src/graph_qwen3.cpp
+++ b/driver/portable/src/graph_qwen3.cpp
@@ -406,14 +406,18 @@ GraphResult build_qwen3_graph(ggml_context* ctx,
                                     rope_beta_fast, rope_beta_slow);
             }
         } else {
+            // Llama3 GGUFs use ggml normal RoPE pairing; NEOX rotates the
+            // wrong dimension pairs and the model attends incoherently.
+            const int rope_type = (h.arch == PieArch::Llama3)
+                ? GGML_ROPE_TYPE_NORMAL : GGML_ROPE_TYPE_NEOX;
             Q = ggml_rope_ext(ctx, Q, in.pos_input, c_rope,
-                              head_dim, GGML_ROPE_TYPE_NEOX, rope_n_ctx_orig,
+                              head_dim, rope_type, rope_n_ctx_orig,
                               layer_rope_theta, rope_freq_scale,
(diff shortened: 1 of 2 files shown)


shsym:

---

Rebased onto current pie/main (past 905498f4 'Add multimodal support to the portable driver'). That commit restructured the qwen3 graph RoPE block into `if (h.use_mrope) { ggml_rope_multi } else { ggml_rope_ext NEOX }`. Resolved by applying this PR's arch-aware `rope_type` (Llama3=NORMAL, else NEOX) inside the new `else` (plain-RoPE) branch — correct since Llama3 never uses M-RoPE. `git range-diff` confirms the net change is identical, only relocated. llama3.rs eom-stop hunk rebased clean. Verified: `cargo test -p pie --lib model::instruct::llama3` 6/6 pass; portable engine builds clean on Metal (arm64). Now MERGEABLE.

PR #428, opened 2026-06-17
fix(inference): never serve a masked decode from a staged speculation fire

shsym:
**Problem:** the pass-level speculation chain extender pre-fires the next decode step before the inferlet sets up the next position's constraints, so every staged fire is built with an empty constraint surface — `logit_masks = []` and `has_user_mask = false`. `entry_matches_request` compared the anchor and spec arrays but neither mask, so a request that carried one was served the unconstrained, fully-causal staged token verbatim — the constraint silently dropped at the unconstrained→constrained transition. Two default-on variants: a grammar / schema / forced-tool **logit** mask (a real model hides it by naturally emitting valid JSON; the dummy driver exposes it as garbage), and a custom **attention** mask (windowed / sink / hierarchical), served fully-causal so it attends to positions the inferlet meant to exclude.

**Fix:** reject a staged hit whenever the request carries either mask, forcing it onto the cold path where the mask is applied. Mirrors the producer-side skip set in `evaluate_request_shape`; the two sites must stay symmetric. Adds regression tests for both mask kinds.

### DIFF
--- a/runtime/src/inference/speculator.rs
+++ b/runtime/src/inference/speculator.rs
@@ -93,6 +93,15 @@ pub(crate) fn entry_matches_request(
     if request.token_ids.len() != 1 {
         return false;
     }
+    // A staged fire is pre-built with an empty constraint surface
+    // (logit_masks = [], fully causal): the next position's grammar /
+    // attention mask isn't known yet. So a request carrying a logit mask
+    // (grammar / schema / forced-tool) or a custom attention mask must
+    // miss and take the cold path where the mask is applied. Mirrors the
+    // producer skip set in `evaluate_request_shape`; keep them symmetric.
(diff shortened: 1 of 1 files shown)


ingim:
Good catch. Do you think there's a common masking case the speculator could reasonably predict ahead of time? For constrained decoding it's impossible, but for hierarchical masks maybe we could?

shsym:
@ingim I can think of two ways:
- (1) the inferlet hands the speculator the mask family (sliding-window / sink / hierarchical) so it can extend the mask exactly per-position, and/or
- (2) a simple heuristic guess: the speculator carries the previous (custom) mask forward by one position.

It can be a separate PR once I test things on Rational's profiles.

PR #429, opened 2026-06-17
feat(inference): per-token logit-bias forward-pass primitive

shsym:
Adds a `logit-bias` method to the `forward-pass` WIT resource — a list of (token_id, bias) pairs added to next-token logits before sampling (after any BRLE logit mask). Enables cross-sibling token penalties for tree-of-thought diverse decoding (RatioThink #693).

Layers: WIT (runtime + SDK + bakery copies), host runtime (`HostForwardPass::logit_bias` + CSR finalize), bridge (`logit_bias_tokens`/`values`/`indptr` appended to ForwardRequest, append-only; C-ABI readers + Desc + view.hpp + round-trip/layout tests), portable driver (apply after mask, before sampler), CUDA driver (bias-aware host-override argmax — same approximation as the existing BRLE-mask override), SDK (`Forward::logit_bias` / `Generator::logit_bias`). Additive and inert when no bias is supplied.

Consumed by RatioThink#276.


ingim:
Adding logit bias is a good idea! 

It looks like this branch may be based on an orphaned commit. It's changing 193 files. Could you merge main into the branch and keep only the logit-bias changes here?

shsym:
It was designed for the Rational app. I will test it on the app first and create a separate PR for the main branch.

shsym:
Re-targeted base from `main` to `pie.app/v1-base-shmem` — this branch is `dbc71826` (the commit the RatioThink app pins, on the shmem fork) + the logit-bias primitive only. Diff is now +303/-11 across 21 files, MERGEABLE/CLEAN. The earlier main-based diff (+5256/-11271) was the shmem-fork divergence, not this change.

PR #430, opened 2026-06-17
Dedupe uniform-top-K sampling reshape into graph_common; key on n_slots

shsym:
## What

The five portable graph builders (`graph_qwen3`, `graph_qwen3_5`, `graph_gemma4`, `graph_phi3_5moe`, `graph_phi3small`) each carried a byte-identical sampling-output tail (greedy `ggml_argmax` / uniform-top-K `top_k`+gather / raw logits). Factor it into a shared `build_sampling_outputs()` in `graph_common`; each builder's tail collapses to one call.

## Invariant guarded

The uniform-top-K gather reshape keys on the slot count `probs->ne[1]` (= n_slots), which is what the per-builder code already did after `dbc71826` fixed the n_req-vs-n_slots bug: pass-level speculation samples more than one slot per request (draft verification), so `n_slots >= n_req`, and a reshape keyed on `n_req` trips GGML's `nelements == ne0*ne1*ne2` assert. The helper derives both `vocab` and `n_slots` from `probs` itself, so the invariant lives in one place instead of five.

## Regression test

`graph_sampling_testhook.cpp` is a test-only C entry that builds the sampling subgraph over a synthetic `[vocab, n_slots]` logits tensor and reports `top_k_probs->ne[1]`. A `#[cfg(test)]` Rust test in the `pie` bin asserts it equals `n_slots`.

- Lives in the bin (not a driver ctest, which never runs in CI) so `cargo test -p pie-server --bin pie` exercises it.
- Mutation-verified: keying the reshape on `n_req` aborts via `GGML_ASSERT(ggml_nelements(a) == ne0*ne1*ne2)` (SIGABRT).
- Release builds GC the unreferenced test symbol — zero production cost.

## Verification

- `cargo test -p pie-server --bin pie` → 2 new tests pass; full suite green.
- App-side: `make test-e2e-http` 202 passed / 1 skipped at this pin.

Base `pie.app/v1-base-shmem`; branched from the app's current pin `2714bd22` so the app bump stays additive.

### DIFF
--- a/driver/portable/CMakeLists.txt
+++ b/driver/portable/CMakeLists.txt
@@ -102,6 +102,10 @@ add_library(pie_driver_portable_lib STATIC
   src/sampler.cpp
   src/host_swap_pool.cpp
   src/adapter.cpp
+  # Test-only entry exercising build_sampling_outputs() from a pie-server
+  # #[cfg(test)] Rust test (driver/portable ctest never runs in CI). The
+  # symbol is GC'd out of release builds — no Rust code references it there.
+  src/graph_sampling_testhook.cpp
 )
 
 target_compile_options(pie_driver_portable_lib PRIVATE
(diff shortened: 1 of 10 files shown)


ingim:

PR #437, opened 2026-06-19
fix(scheduler): relax AdaptivePolicy cohort watermark instead of ratcheting monotonically

### LINKED ISSUE #436: Pass-level speculation under batch saturation regresses long-context concurrent (MoE) decode 2–4×
## Summary

Pass-level speculation (`[model.scheduler].speculation_depth = 1`, the default) appears to be
**working as intended** for the case it targets: it definitively *helps* single-stream /
short-context decode. However, it runs **unconditionally** — it does not back off when the
scheduler batch is already saturated. Under sustained concurrency with **long contexts**, the
pre-fired speculative steps inflate each batch, and on a memory-bandwidth-bound long-context
**MoE** model the larger batches cost more per fire than the round-trips they save. Net result:
**2–4× lower decode throughput** vs `speculation_depth = 0`, on identical work.

This is not a correctness issue — outputs are fine, the speculation hit rate is ~99%, and
scheduling is measured to be *fair* (not starvation). It's a performance regression from batch
inflation under saturation, and the proposed improvement is to **gate speculation on spare
capacity**.

## Environment

- Branch `pie.app/v1-base-shmem` @ `14f09d9d` (`runtime/src/inference/speculator.rs` also present on `main`).
- Driver `portable`, built `PIE_PORTABLE_CUDA=1 PIE_PORTABLE_CUDA_ARCH=89`. GPU NVIDIA L40S (CC 8.9), CUDA 12.8.
- Model `Qwen/Qwen3-30B-A3B-GGUF` (Q4_K_M) — MoE, 8 of 128 experts active per token.
- Defaults: `speculation_depth = 1`, `batch_policy = "adaptive"`. Driver opts: `max_forward_tokens = 16384`, `total_pages = 2048`, `kv_page_size = 32`, `max_forward_requests = 32`.

## Reproduction (controlled A/B)

Identical workload run twice — **only `speculation_depth` differs** (0 vs 1). Same tasks, seed,
concurrency, max_tokens, engine build; fresh engine restart per arm; equal 25-minute budget.
Workload: [tau2-bench](https://github.com/sierra-research/tau2-bench) `airline`,
`--num-tasks 12 --max-concurrency 4 --seed 10`, `max_tokens = 16384`, served through an
OpenAI-compatible HTTP inferlet. (The trigger is concurrent long generations, not anything
benchmark-specific.)

| metric (equal 25-min budget) | `speculation_depth=0` | `speculation_depth=1` |
|---|---|---|
| effective decode throughput | **23.6 tok/s** | **6.1 tok/s** |
| total tokens generated | 54,897 | 9,401 |
| tasks completed / 12 | **12** | **3** |
| completed LLM calls | 76 | 12 |
| min completed-call tok/s | 15.2 | **0.29** |
| completed calls &lt; 5 tok/s | 0 % | 8.3 % |

A first A/B (without verbose logging) showed the same direction at smaller magnitude (off 24.8
tok/s 9/12 tasks; on 11.3 tok/s 7/12). Magnitude varies run-to-run with task mix (some tasks
generate far longer than others), but `depth=0` is consistently ~2–4× faster and completes more
tasks.

## Evidence: saturation + large batches (per-fire engine trace)

I added a per-batch trace line in `BatchScheduler::run` (`[pie-sched-ctx] t_us=.. ctxs=[..]`,
gated on an env var) and compared the two arms over the same workload:

| per-fire metric | `depth=0` | `depth=1` |
|---|---|---|
| batch fire rate | 24.2 / s | 16.3 / s |
| inter-fire gap, median | 35.8 ms | **50.7 ms (+42 %)** |
| inter-fire gap, p95 | 60 ms | **134 ms (+123 %)** |
| batch-size distribution | mostly **4** | mostly **4–7 (larger)** |
| within-window fairness (max/min fires among co-active contexts) | median 2.3× | median 1.0× |

Reading:
- **The batch is already saturated** under concurrency (fires back-to-back, GPU ~76–94 % util) —
  there is no idle latency for pre-firing to hide.
- **Speculation enlarges the batch** (size 4 → 5–7) by adding pre-fired steps, and each fire then
  takes ~42 % longer (median) with a 2× worse tail.
- **It is not starvation** — co-active contexts receive *equal* fires (fairness ratio 1.0×, even
  more even than `depth=0`). The slow calls come from the heavy fire-time tail, not from any
  context being skipped.

## Why this happens (mechanism)

On a memory-bandwidth-bound decode, a bigger batch only pays off when it amortizes a weight read
that's happening anyway. Two effects defeat that here:

1. **Long context → expensive attention per added token.** The speculative token must attend over
   the full KV cache (10k–16k tokens in this workload), so each extra batch token is costly. This
   part is architecture-agnostic.
2. **MoE → extra experts loaded.** Different tokens route to different experts (8 of 128 here), so
   a larger, more-diverse batch can force loading **more distinct experts**, adding real
   feed-forward memory traffic. On a dense model the extra tokens would mostly ride the single
   weight read for free; on MoE they don't. (Not yet tested on a dense model — expectation is the
   regression would be milder there.)

Because the GPU is already saturated under concurrency, the round-trips speculation hides are not
on the critical path, so the added per-fire cost is a net loss.

Scope: short/bounded loads do **not** regress — symmetric/staggered concurrent requests with
bounded `max_tokens` stay fair at 24–36 tok/s under both settings, single stream is ~125 tok/s
with speculation on, and a faithful tau2-shaped sustained churn that capped `max_tokens` low
stayed fair. A clean dose–response on generation length: ~1.5k-token contexts fair → 8k-token
~1.9× spread → 16k-token (tau2) 2–4×. The regression specifically needs **long context +
saturation**.

## Suggested improvement

Gate pass-level speculation on **spare batch / GPU capacity** — only pre-fire when the current
batch is below the occupancy/token limits (i.e. there is idle to hide), and skip when the
scheduler is saturated. Today `speculator::start_chain` (`runtime/src/inference.rs`) tags every
cold submit as a self-perpetuating `Completion::Chain`, and the chain re-extension in
`process_chain_job` (`runtime/src/inference/speculator.rs`) gates only on a per-context queue
depth (`should_extend = queued &lt;= max_queue_depth`) with no global load signal. A minimal change
is to AND in a saturation flag (e.g. recent batch occupancy near `max_forward_requests`, or a
persistently non-empty scheduler queue / `next_pending`) maintained by `BatchScheduler::run`
(`runtime/src/inference/scheduler.rs`). Under saturation the chain stops extending and falls back
to the normal cold/actor path, recovering `depth=0` throughput, while light/single-stream load
keeps the speculation win.

## Workaround

Set `speculation_depth = 0` under `[model.scheduler]` for concurrent long-context workloads.

evanc7007:
## Summary

`AdaptivePolicy::fired_high_water` only ever ratchets **up** and never decays. Once a transient
burst lifts it above the steady-state cohort size, every later fire whose batch lands *below*
that peak fails the rule-3 firing test (`B >= fired_high_water`) and is parked on the rule-4
`last_latency` watchdog before it goes out — adding a full ~one-fire-time stall to a large
fraction of fires for the rest of the engine's lifetime.

This PR makes `fired_high_water` track the **recent** cohort instead of an all-time peak: it
still ratchets up immediately on a meet-or-exceed fire, but relaxes by one (floored at the
observed size) on a below-watermark fire, so it follows the live cohort back down. Fixes #436 

## Why the current behavior is a bug

The watermark exists to avoid firing a half-empty batch when more peers are about to re-submit.
But because it never decays, *any* event that lifts it above the current steady state poisons
every subsequent sub-peak fire:

- **Concurrency decline** (no speculation needed): as in-flight requests finish, the active
  cohort falls from its peak (say 6) through 5, 4, 3… Each sub-peak cohort now fails rule 3 and
  waits out the watchdog, even though no more peers are coming.
- **Pass-level speculation**: speculative pre-fires transiently inflate a batch (e.g. 4 → 6–7),
  lifting the watermark; the normal cohort of ~4–5 is then permanently "below peak" and stalls.

So this is a general scheduler-policy defect, not a speculation-specific one — speculation is
just one trigger. (It surfaced first under speculation because that re-triggers the condition
continuously; see the speculation throughput discussion in #434's companion thread.)

## Reproduction & evidence

Workload: `tau2-bench` airline, `--num-tasks 12 --max-concurrency 6 --seed 10`,
`max_tokens=16384`, served via an OpenAI-compatible inferlet; 25-min budget per arm; engine
built `PIE_PORTABLE_CUDA=1 PIE_PORTABLE_CUDA_ARCH=89`, NVIDIA L40S, Qwen3-30B-A3B-GGUF (Q4_K_M).
Per-fire scheduler trace parsed into inter-fire gap by batch size; "stall %" = fires whose
preceding gap exceeds 2× the dominant (at-watermark) cohort's median.

| arm | watermark | stall % | inter-fire p95 |
|---|---|---|---|
| spec-off (no speculation) | monotonic (today) | **37.5%** | 104 ms |
| spec-on | monotonic (today) | 23.9–25.2% | 117–127 ms |
| spec-on | **relaxing (this PR)** | **0.2–0.6%** | **55–69 ms** |

Note the spec-**off** control already shows 37.5 % stalled fires purely from concurrency
decline — the bug reproduces with speculation entirely disabled. With the fix, the sub-peak
cohorts that previously sat at 100–125 ms (batch sizes 4–5 while the watermark was pinned at 6)
drop to ~40–65 ms, and the relaxing-watermark spec-on arms end up *cleaner than the monotonic
spec-off control*. Result reproduced in both verbose and quiet logging configurations.

## The change

`runtime/src/inference/adaptive_policy.rs` only:

```rust
fn relax_high_water(current: usize, fired_size: usize) -> usize {
    if fired_size >= current { fired_size } else { current.saturating_sub(1) }
}
// on_fired: self.fired_high_water = relax_high_water(self.fired_high_water, fired_size);
```

Decay-by-one (rather than e.g. snapping straight to the live size, or a sliding-window max) was
chosen against the recorded fire-size sequences: a window-max doesn't relax when large batches
recur, while snapping to the size of a single fire is too eager; stepping down by one removed
~99.7 % of the watchdog stalls in replay and reacts within 1–2 fires, while ignoring a lone
anomalously small fire. (On the observed traces the cohort never fell by more than one between
consecutive fires, so a snap-to-size variant would have been indistinguishable here.)

**Behavior preserved:** cold-start (`fired_high_water == 0`), the structural cap, and the
`last_latency` watchdog are unchanged. A uniform cohort keeps the watermark exactly at the
cohort size (verified by test), so steady-state batching and the existing coalescing benefit
are untouched — only the never-decaying tail behavior changes.

## Tests

`cargo test -p pie -- adaptive relax_high_water`: ratchet-up-on-exceed, relax-after-burst
(7 → settles at the live cohort of 5), steady-cohort stability (uniform load stays at the
cohort size, never inflates or under-fires), and the `relax_high_water` floor cases.

## Summary by CodeRabbit

* **Improvements**
  * Adaptive policy watermark behavior enhanced to dynamically respond to workload changes with both upward and downward adjustments, providing improved flexibility in system adaptation.
* **Documentation**
  * Updated documentation reflecting the new adaptive policy watermark adjustment patterns.
* **Tests**
  * Expanded unit tests to validate watermark behavior under various workload scenarios and adjustment patterns.

### DIFF
--- a/runtime/src/inference/adaptive_policy.rs
+++ b/runtime/src/inference/adaptive_policy.rs
@@ -32,8 +32,8 @@ use super::scheduler::{Decision, SchedulingPolicy};
 //   1. `B >= max_forward_requests`  — structural driver limit.
 //   2. `fired_high_water == 0`      — no historical fire yet; fire
 //                                     whatever we have (cold start).
-//   3. `B >= fired_high_water`      — we've matched the largest cohort
-//                                     ever fired; firing more won't
+//   3. `B >= fired_high_water`      — we've matched the recent cohort
+//                                     high-water; firing more won't
 //                                     produce a bigger batch from
 //                                     existing peers.
 //   4. `elapsed >= last_latency`    — deadlock watchdog: the cohort
@@ -52,11 +52,44 @@ use super::scheduler::{Decision, SchedulingPolicy};
 // `fired_high_water` is the only firing-rule concept; `last_latency`
 // is a watchdog to prevent indefinite parking, not a steady-state
 // trigger.
+//
+// `fired_high_water` tracks the *recent* cohort high-water, not an
+// all-time maximum: it ratchets up immediately when a fire meets or
+// exceeds it, and relaxes by one on any below-watermark fire (see
+// `relax_high_water`). A monotonic, never-decaying watermark is a bug:
+// once a transient burst lifts it (e.g. a one-off larger cohort, or
+// pass-level speculation inflating a batch), every later steady-state
+// cohort that lands below the peak fails rule 3 and needlessly waits
+// out the rule-4 watchdog. Relaxing the watermark lets it follow the
+// live cohort back down so those fires go out immediately.
+
+/// Evolve `fired_high_water` after a fire of `fired_size`.
+///
+/// Ratchets up immediately when the fire meets or exceeds the current
+/// watermark; otherwise relaxes by exactly one toward the live cohort.
+/// Stepping down by one (rather than snapping straight to `fired_size`)
+/// is deliberate: it ignores a single anomalously small fire but tracks
+/// a sustained drop within a couple of fires, which matches observed
+/// cohort dynamics (concurrency rarely falls by more than one between
+/// consecutive fires).
+///
+/// Free function (not a method) so the firing-rule arithmetic is unit
+/// testable in isolation.
+#[inline]
+fn relax_high_water(current: usize, fired_size: usize) -> usize {
+    if fired_size >= current {
+        fired_size
+    } else {
+        current.saturating_sub(1)
+    }
+}
 
 pub(super) struct AdaptivePolicy {
     max_forward_requests: usize,
-    /// Largest batch size that has fired. Monotonic; only ratchets
-    /// upward. The single firing-rule concept.
+    /// Recent cohort high-water — the single firing-rule concept.
+    /// Ratchets up on a meet-or-exceed fire and relaxes by one on a
+    /// below-watermark fire (see `relax_high_water`), so it tracks the
+    /// current steady-state cohort rather than an all-time peak.
     fired_high_water: usize,
     /// Previous fire's compute time, in seconds. Used solely as the
     /// deadlock watchdog (rule 4). Zero until the second fire
@@ -97,9 +130,7 @@ impl SchedulingPolicy for AdaptivePolicy {
 
     fn on_fired(&mut self, fired_size: usize) {
         self.batch_start_time = None;
-        if fired_size > self.fired_high_water {
-            self.fired_high_water = fired_size;
-        }
+        self.fired_high_water = relax_high_water(self.fired_high_water, fired_size);
     }
 
     fn decide(&mut self, current_forward_requests: usize) -> Decision {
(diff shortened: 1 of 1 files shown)


ingim:
Great work, @evanc7007 !

PR #443, opened 2026-06-22
feat(sdk): allow Generator context checkpoints

### LINKED ISSUE #405: Generator holds &mut Context, so checkpointing mid-generation requires dropping the generator (tearing down the decode session)
## Summary

  There is no way to snapshot or fork a `Context` *during* an active generation
  session. `Context::generate()` borrows the context mutably for the entire
  lifetime of the returned `Generator`, but `Context::snapshot()` / `fork()` take
  `&self`, and `Generator` exposes no accessor to its context. The only way to
  checkpoint is to `drop` the `Generator` first — which ends the decode session.

  For loops that want to checkpoint at a token boundary and *keep generating*
  (e.g. async tool-calling with mid-stream result injection), this forces a
  session teardown/rebuild at every checkpoint, which has real behavioral cost
  (see "Impact").

  ## Where

  - `sdk/rust/inferlet/src/generation.rs:55` — `pub struct Generator<'ctx> { ctx: &'ctx mut Context, … }` (field is private; no accessor)
  - `sdk/rust/inferlet/src/context.rs:428` — `pub fn generate(&mut self, …) -> Generator<'_>` (takes `&mut self`)
  - `sdk/rust/inferlet/src/context.rs:161` — `pub fn snapshot(&self) -> Result<String>`
  - `sdk/rust/inferlet/src/context.rs:140` — `pub fn fork(&self) -> Result<Self>`

  ## Repro

  ```rust
  let mut g = ctx.generate(sampler);          // &mut ctx borrowed for g's lifetime
  while let Some(step) = g.next()? {
      let out = step.execute().await?;
      if at_checkpoint_boundary(&out) {
          let name = ctx.snapshot()?;          // ❌ cannot borrow `ctx` as immutable
                                               //    while borrowed as mutable by `g`
          // ...want to keep using `g` here...
      }
  }

  Generator offers no g.snapshot() / g.context() either, so the only
  compile-able workaround is:

  drop(g);                  // end the &mut borrow (and the decode session)
  let name = ctx.snapshot()?;
  let mut g = ctx.generate(sampler);   // rebuild from scratch

  Impact

  In an async tool-calling inferlet, I checkpoint on each [CALL] boundary so a
  later trap can Recompute/restore. Because checkpointing requires dropping the
  generator, the decode loop is torn down and restarted once per call. That
  restart exits before the bottom-of-loop mid-stream injection check, so
  completed call results can't be spliced into the same decode stream — they're
  deferred to the next trap. Net effect: you can have per-call checkpointing OR
  full mid-stream overlap, but not both, purely because of this borrow
  constraint, not because of any runtime limitation.

  Proposed fix

  Add &self passthroughs on Generator that reborrow self.ctx (both are
  &self on Context, so this is sound):

  impl<'ctx> Generator<'ctx> {
      /// Snapshot the underlying context without ending the session.
      pub fn snapshot(&self) -> Result<String> { self.ctx.snapshot() }

      /// Fork the underlying context (copy-on-write pages + buffer).
      pub fn fork(&self) -> Result<Context> { self.ctx.fork() }
  }

  Or, more generally, a shared accessor so callers can reach any &self method
  (buffer(), seq_len, etc.):

  impl<'ctx> Generator<'ctx> {
      pub fn context(&self) -> &Context { self.ctx }
  }

  Note for a correct restore point: snapshot() saves only the runtime inner
  state and does not capture the SDK-side Context::buffer() (the last
  sampled token pending flush), so callers currently have to stash buffer()
  alongside the snapshot name and replay it. fork() does clone the buffer
  (context.rs:147), so a Generator::fork() passthrough would be the cleaner
  complete-restore primitive — but it's gated by the same borrow today.

  Workaround

  Drop the generator, snapshot, and rebuild (as above) — at the cost of a decode
  session restart per checkpoint.

leo-lilinxiao:
## Summary

Adds `Generator` passthroughs for context checkpoint operations while a generation session is still active:

- `Generator::context()` for read-only context inspection
- `Generator::fork()` for an in-memory checkpoint that also clones the SDK-side token buffer
- `Generator::save()` / `Generator::snapshot()` for host-side runtime snapshots

The generate test inferlet now exercises `g.context()` and `g.fork()?` after the first decode step, while the same `Generator` continues running.

## Why

`Context::generate()` borrows the context mutably for the lifetime of the returned `Generator`, so callers previously had to drop and rebuild the generator before checkpointing. That makes mid-generation checkpoint/restore patterns awkward, especially for tool-calling or recovery loops that want to checkpoint at token boundaries without tearing down the decode session.

Fixes #405.

## Verification

- `git diff --check main..feat/generator-checkpointing`
- `cargo build --manifest-path runtime/tests/inferlets/Cargo.toml --target wasm32-wasip2 -p generate`
- `cargo check -p pie`
- `cargo test -p pie --lib` — 433 passed, 0 failed, 2 ignored

## Summary by CodeRabbit

* **New Features**
  * Added the ability to save and fork generation context during inference, allowing users to preserve state mid-generation without interrupting the current session.

### DIFF
--- a/runtime/tests/inferlets/generate/src/lib.rs
+++ b/runtime/tests/inferlets/generate/src/lib.rs
@@ -34,10 +34,21 @@ async fn main(_input: String) -> Result<String> {
         .max_tokens(max_tokens);
 
     let mut generated = Vec::new();
+    let mut checkpointed_mid_generation = false;
     while let Some(step) = g.next()? {
         let out = step.execute().await?;
         if out.tokens.is_empty() {
             continue;
+        }
+        if !checkpointed_mid_generation {
+            let buffered_after_step = g.context().buffer().len();
+            let checkpoint = g.fork()?;
+            eprintln!(
+                "[GENERATE] forked checkpoint mid-generation with {} buffered tokens",
+                buffered_after_step
+            );
+            drop(checkpoint);
+            checkpointed_mid_generation = true;
         }
         eprintln!("[GENERATE] got tokens: {:?}", out.tokens);
         generated.extend(out.tokens.iter().copied());
--- a/sdk/rust/inferlet/src/generation.rs
+++ b/sdk/rust/inferlet/src/generation.rs
@@ -232,6 +232,47 @@ impl<'ctx> Generator<'ctx> {
         ProbeHandle::new(slot)
     }
 
+    // ── Context checkpoints ──────────────────────────────────────────
+
+    /// Borrow the underlying context without ending the generation
+    /// session. This is useful for read-only context operations such as
+    /// inspecting `buffer()`, page counts, or scheduler state while the
+    /// generator still owns the mutable borrow.
+    pub fn context(&self) -> &Context {
+        &*self.ctx
+    }
+
+    /// Fork the underlying context without ending the generation session.
+    ///
+    /// Unlike [`snapshot`](Self::snapshot), `Context::fork` also clones the
+    /// SDK-side token buffer, including any sampled token staged for the
+    /// next decode step. Prefer this for a complete in-memory checkpoint.
+    pub fn fork(&self) -> Result<Context> {
+        self.ctx.fork()
+    }
+
+    /// Save the underlying runtime context under `name` without ending the
+    /// generation session.
+    ///
+    /// This saves the host-side context state. It does not persist the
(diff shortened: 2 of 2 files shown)


ingim:
Thanks for submitting the fix!

PR #444, opened 2026-06-22
fix(runtime): enforce network policy for wasi:http

leo-lilinxiao:
## Problem

`wasi:http` runs through Wasmtime's host HTTP stack, not the `wasi:sockets` connect hook. Pie already dropped the HTTP linker binding when `allow_network = false`, but with `allow_network = true` a restricted `network_allowed_hosts` policy still did not constrain outbound HTTP requests.

That meant an inferlet with a socket allowlist could still use `wasi:http` to reach hosts outside the allowlist.

## Fix

Override `WasiHttpView::send_request` in `InstanceState` and check the outgoing request authority before handing it to Wasmtime's default HTTP client.

- `allow_network = false`: unchanged; `wasi:http` is not linked.
- `network_allowed_hosts = ["*"]`: preserves legacy unrestricted hostname behavior.
- restricted allowlists: outbound HTTP must use an IP-literal authority and pass the same CIDR/port rules as `wasi:sockets`.

Restricted mode deliberately rejects hostname authorities instead of resolving them in policy code, avoiding a DNS time-of-check/time-of-use gap before the default HTTP client opens the connection. Denied requests return `ConnectionRefused` and log only scheme/host/port, not the full URI.

The runtime config comments and generated config template are updated to describe the HTTP behavior.

## Verification

- `git diff --check main..feat/wasi-http-network-policy`
- `cargo check -p pie`
- `cargo test -p pie --lib policy::tests` — 36 passed, 0 failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Network allowlist policies now consistently apply to outbound HTTP requests, matching `wasi:sockets` behavior.
  * HTTP requests in restricted policies are limited to IP-literal authorities to prevent DNS timing-of-check/time-of-use vulnerabilities.

* **Documentation**
  * Updated configuration guidance to clarify network policy enforcement across both socket and HTTP interfaces.

### DIFF
--- a/runtime/src/bootstrap.rs
+++ b/runtime/src/bootstrap.rs
@@ -81,9 +81,9 @@ pub struct RuntimeConfig {
     /// linker binding is dropped entirely.
     pub allow_network: bool,
     /// Allowlist of `cidr[:port]` / `cidr:lo-hi`. `["*"]` = no
-    /// restriction. NOTE: only filters `wasi:sockets`; `wasi:http`
-    /// bypasses the per-socket hook. Set `allow_network = false` for
-    /// tight outbound HTTP control.
+    /// restriction. Applies to both `wasi:sockets` and outbound
+    /// `wasi:http`; restricted HTTP policies accept IP-literal authorities
+    /// only to avoid DNS time-of-check/time-of-use gaps.
     pub network_allowed_hosts: Vec<String>,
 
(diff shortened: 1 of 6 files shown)


ingim:
Great find @leo-lilinxiao, thank you for submitting the fix

PR #491, opened 2026-07-22
Fix AWQ->Marlin zero-point repack: add the 8-wide interleave

shsym:
The cuda_native AWQ-INT4 zero-point converter (awq_qzero_to_marlin_w4_kernel in driver/cuda/src/kernels/dtype_cast.cu) applied the 64-wide Marlin scale permutation but omitted the subsequent 8-wide column interleave [0,2,4,6,1,3,5,7] that marlin_zero_points performs before packing. As a result every group's zero-points were mis-permuted, so dequantized weights (q - z) * s were wrong on every output channel and decode collapsed to a single repeated token for asymmetric AWQ checkpoints (e.g. Qwen2.5-72B-Instruct-AWQ). The qweight and scale paths were already correct.

The one-line kernel change emits output nibble slot j from logical column n8_out*8 + interleave[j] instead of the linear n8_out*8 + j, matching vLLM's awq_to_marlin_zero_points. A new direct-kernel CUDA regression test (driver/cuda/tests/test_awq_marlin_zero_points.cu, registered as the awq_marlin_zero_points CTest target) runs the actual kernel on device and compares bit-for-bit against an independent host reference of the full undo-interleave -> scale_perm -> interleave -> repack pipeline across several geometries including the group-size-128 case, and asserts the pre-fix linear mapping differs. The existing GPTQ-symmetric path never calls this kernel and is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CUDA AWQ quantized zero-point conversion for Marlin format by applying the required 8-wide interleave after the existing scale permutation.
  * Updated int4 zero-point repacking to match the expected output layout.

* **Tests**
  * Added a new CUDA regression test (`awq_marlin_zero_points`) that compares the device zero-point lowering against an independent CPU reference (and validates the interleave is actually applied).

### DIFF
--- a/driver/cuda/CMakeLists.txt
+++ b/driver/cuda/CMakeLists.txt
@@ -569,6 +569,15 @@ target_compile_options(test_awq_repack_source_ordering PRIVATE
   $<$<COMPILE_LANGUAGE:CUDA>: --extended-lambda --expt-relaxed-constexpr >)
 target_link_libraries(test_awq_repack_source_ordering PRIVATE CUDA::cudart)
 
+add_executable(test_awq_marlin_zero_points
+  tests/test_awq_marlin_zero_points.cu
+  src/kernels/dtype_cast.cu)
+target_include_directories(test_awq_marlin_zero_points PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}/src)
+target_compile_options(test_awq_marlin_zero_points PRIVATE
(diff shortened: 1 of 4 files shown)


shsym:

---

@coderabbitai review

shsym:

---

@coderabbitai review

shsym:
@coderabbitai review

PR #492, opened 2026-07-24
Order transient AWQ repack/cast source uploads on stream 0

shsym:
## Problem

Loading an AWQ checkpoint through the `cuda_native` driver can produce corrupt
weights (incoherent decode) or a pre-readiness SIGSEGV. The AWQ repack/cast
transient-source uploads in `TranscodeEngine` use the deferred, cross-stream
`copy_engine_.queue()`, which only executes at the next `flush()` — after the
repack/cast kernels have already launched on stream 0, and after the scratch
`DeviceTensor` has been freed. The kernels therefore read uninitialized device
memory (garbage-built GPTQ/Marlin weights), and the deferred reader-lane copy
later writes into freed device memory (a SIGSEGV inside libcuda). The timing/
layout sensitivity makes it intermittent.

## Fix

Order both `materialize_repack_source` and `cast_tile_map` uploads on stream 0 via
`copy_engine_.queue_on_stream(..., /*stream=*/0)`, mirroring the existing
encode/fp8 sibling paths. Stream-0 ordering guarantees each kernel sees its fully
uploaded source, and — because `cudaFree` synchronizes the device — keeps the
scratch alive until the kernel completes.

## Test

`test_awq_repack_source_ordering` (CTest `awq_repack_source_ordering`) drives the
real `launch_awq_qweight_to_gptq_w4` kernel: an ordered upload matches an
independent host reference bit-for-bit, while a misordered upload provably reads
the poison the scratch was left with — guarding against a test that would pass
even if ordering were ignored.

## Verification

On an A100 with a Qwen2.5-72B-Instruct-AWQ load: compute-sanitizer `initcheck`
reports zero uninitialized `__global__` reads across the full weight load (versus
reads attributed to `awq_qweight_to_gptq_w4_kernel` before the fix), and the model
serves crash-free across repeated loads.

## Note

This fixes the SIGSEGV and the uninitialized-read weight corruption. Full
72B-AWQ decode coherence additionally requires the zero-point interleave fix in
#491 (an orthogonal change to the qzero→Marlin repack); the two are independent
and both required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA stream-ordered memory transfers so transient scratch-buffer uploads occur before repack/cast kernels, preventing kernels from reading stale data.
* **Tests**
  * Added a CUDA unit test that deterministically verifies AWQ repack source ordering: the correctly ordered run matches the host reference, while a deliberate mis-order reproduces the expected “poison” outcome.

### DIFF
--- a/driver/cuda/CMakeLists.txt
+++ b/driver/cuda/CMakeLists.txt
@@ -560,6 +560,15 @@ target_compile_options(test_transcode_fused PRIVATE
   $<$<COMPILE_LANGUAGE:CUDA>: --extended-lambda --expt-relaxed-constexpr >)
 target_link_libraries(test_transcode_fused PRIVATE CUDA::cudart)
 
+add_executable(test_awq_repack_source_ordering
+  tests/test_awq_repack_source_ordering.cu
+  src/kernels/dtype_cast.cu)
+target_include_directories(test_awq_repack_source_ordering PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}/src)
+target_compile_options(test_awq_repack_source_ordering PRIVATE
+  $<$<COMPILE_LANGUAGE:CUDA>: --extended-lambda --expt-relaxed-constexpr >)
+target_link_libraries(test_awq_repack_source_ordering PRIVATE CUDA::cudart)
+
 enable_testing()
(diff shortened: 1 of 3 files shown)


shsym:
@coderabbitai review

PR #508, opened 2026-08-04
feat(context): report what opening a snapshot cost

Leonana69:
> Supersedes #504, which GitHub auto-closed when the head fork was recreated. Same commit, unchanged.

## Problem

Eviction *suspends* a context rather than deleting it, and leaves the name resolvable. So `open` succeeds on an evicted snapshot and silently replays the missing prefix through real forward passes.

From the guest's side those two cases are indistinguishable — `Context::open` returns `Result<Self>`, which has no third state for *"found, but I rebuilt it"*. A caller doing prefix caching can report "reused 800 tokens" when every one of them was re-prefilled, and has no way to know.

## The runtime already computes this

`ContextManager::fork` derives `prefix_len` from the GPU page store and a `has_replay` flag from the replay spawners, uses them to set `State::Pinned` / `pending_replay`, and then returns a bare `Ok(new_id)` — dropping both. `take_inner` does the same. This returns them instead of discarding them.

## API

Additive. `open` and `take` keep their exact signatures, so existing guests are unaffected:

```wit
record open-report {
    resident-prefix-pages: u32,
    replayed-pages: u32,
    rs-replayed: bool,
}

open-with-report: static func(model: borrow<model>, name: string)
    -> result<tuple<context, open-report>, error>;
take-with-report: static func(model: borrow<model>, name: string)
    -> result<tuple<context, open-report>, error>;
```

Internally `fork`/`take` delegate to new `fork_with_report`/`take_with_report` and drop the report, so no existing call site changed.

## Why pages, and why `rs-replayed` is separate

"Was it resident" is not a yes/no question, so a boolean would be misleading:

- Residency is per page-hash against the content-addressed trie. A suspended context can fork with **nothing** replayed when another live context still holds its pages.
- Recurrent-state replay (`spawn_full_rs_replay_pass`) runs on a path independent of KV-page replay. A linear-attention model can replay its entire recurrent state while `replayed-pages` is zero — hence the separate flag.

## Notes for reviewers

- **WIT copies were synced by hand.** `runtime/wit/deps/core`, `sdk/rust/inferlet/wit/deps/core`, and `sdk/tools/bakery/.../wit/deps/core` each hold their own copy, and nothing in the repo invokes `wit-deps`. Happy to change the approach if there's a preferred workflow. (Unrelated observation: `runtime/wit/zo/wit/deps/core/context.wit` appears to have already drifted — it's missing the `suspend`/`bid` block. Left alone here.)
- **Python and JavaScript bindings are checked in** and still need regenerating via `componentize-py` and `npm run generate-bindings`. Not done here since I can't verify those toolchains.

## Verification

`cargo check` clean; 439 lib tests, `rs_cache`, and all 9 `contention` tests pass. The guest SDK compiles for `wasm32-wasip2`, which exercises the new WIT end to end through `wit-bindgen`.

Since opening the original, this has also been exercised against a real engine (Qwen3.5-9B) by a downstream consumer: a cold open reports no data, a warm one reports `resident-prefix-pages=1 replayed-pages=0`, and the numbers move as expected across a chat → tree-of-thought → chat sequence.

Independent of #507 — that one is a scheduler bug fix, this is a new surface. Neither depends on the other.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added optional context-opening operations that return resource usage details.
  * Reports now show resident pages, replayed pages, and whether recurrent state was replayed.
  * Added reporting support across the runtime, Rust SDK, and interface definitions.
  * Existing context-opening and retrieval operations remain compatible and unchanged in behavior.


Leonana69:
Added `runtime/tests/open_report.rs` — a direct answer to the question this PR should be judged on: *does the report actually detect a replay, or does it just always say zero?*

It asserts both directions against a forced eviction:

```
resident -> OpenReport { resident_prefix_pages: 1, replayed_pages: 0 }
evicted  -> OpenReport { resident_prefix_pages: 0, replayed_pages: 1 }
```

The inversion is the point. `open` returns `Ok` in both cases and any name-level counter is identical across them, so without the evicted assertion the report could be a constant and every consumer would still look green.

Two things worth flagging for reviewers, because they surprised me:

- **Releasing a snapshot's refcount does not make its pages non-resident.** Pages are content-addressed and shared through the trie, so they survive while any context references them — and even unreferenced pages linger until an allocation reclaims the space. The test drops every reference and then churns the pool.
- **Fork sharing makes this hard to provoke from outside.** Forking N contexts over a common prefix costs that prefix once, not N times. I first tried to observe a replay end-to-end through a real 9B model and could not: either no eviction fired, or the budget dropped below what a single forward pass needs and the engine refused to start. That is why this landed as a runtime test — the mock env reproduces it deterministically in ~10ms.

Test counts in the PR description are now 439 lib + 2 `open_report` + `rs_cache` + 9 `contention`, all passing.

PR #513, opened 2026-08-04
rng-contract: drop the shift width from the magic set, it detects nothing

shsym:
The ownership guard was failing on `driver/metal/tests/llama_numerics_test.cpp`, and the
reported reason was wrong. This corrects the guard, not the file it accused.

## Nothing is transcribed there

Scanning that file for each needle individually, under the test's own normalization
(lowercase, underscores and whitespace stripped):

| needle | present in llama_numerics_test.cpp |
| --- | --- |
| splitmix multiplier 1 | no |
| splitmix multiplier 2 | no |
| golden-ratio stride | no |
| ambient mask | no |
| scale divisor | no |
| right-shift width | **yes**, one hit, line 124 |

The hit is inside `hashf`, whose multipliers are the MurmurHash3 finalizer words and which
maps to `[-1, 1)` using a different divisor. It shares no constant with the contract. There
was no drift, and nothing to single-source.

## Why the needle is the defect

Every other entry in the set is a constant the contract *chose* — re-typing one is real
evidence. A right shift by 40 is not a constant. It is what taking the top 24 bits of a
64-bit hash forces anyone to write.

It also had no detection power to lose. In `ptir_rng_hash_uniform` that shift sits three
lines from the stride and the scale divisor, and each of the four declared homes carries all
of them together, so a genuine transcription trips needles that *are* constants regardless:

| file | mult1 | mult2 | stride | mask | shift | divisor |
| --- | --- | --- | --- | --- | --- | --- |
| `compiler/ir/src/rng.rs` | Y | Y | Y | Y | – | – |
| `compiler/codegen/include/rng_contract.generated.h` | Y | Y | Y | Y | Y | Y |
| `compiler/codegen/include/ptir_rng.generated.metal` | Y | Y | Y | Y | Y | Y |
| `driver/metal/src/kernels/ptir_rng.generated.metal` | Y | Y | Y | Y | Y | Y |
| `driver/metal/tests/llama_numerics_test.cpp` | – | – | – | – | **Y** | – |

The shift never appears alone in a real transcription, and appears alone in the only false
positive. A needle that catches nothing the others miss while firing on unrelated
hash-to-float code is not a guard rail.

## Why not an exemption, and why not editing the Metal test

`unrelated_stride_users` / `unrelated_mask_users` exist for needles that *do* carry detection
power and also collide coincidentally — the golden-ratio word and the ambient mask are
contract constants that legitimately appear in a splitmix id generator and a boost-style
`hash_combine`. Exempting the Metal test would keep a needle with no upside and leave the
same trap for the next hash-to-float anyone writes. Adding it to `owners` would be worse: that
list means "this file is a declared home for the contract", and a fifth home legitimizes the
duplication the test exists to prevent. Editing `hashf` would bend correct code to satisfy a
bad check.

## The whole class, not one instance

The guard `panic!`s on the first offender, so a fix can uncover the next. A repo-wide scan of
all **1286** files the walker visits, for all six needles, confirms this is the complete set:
after this change nothing outside the four declared homes matches any needle. The scale
divisor — what actually pins the uniform conversion — matches in three files, all owners.

## Verification

- `cargo test -p pie-compiler-tests` in full: 16 test binaries, all green. `rng_contract`
  goes from 5 passed / 1 failed to **6 passed / 0 failed**.
- `cargo test --workspace --exclude pie-server-py --exclude pie-worker`: **exit 0**, and it
  now runs `pie_engine`'s unit tests. That step used to abort at `pie-compiler-tests`, which
  is why the engine suite was never reached in that job.
- `cargo clippy --no-deps --all-targets -p pie-compiler-tests -- -D warnings`: clean.
- `cargo fmt --check -p pie-compiler-tests`: clean.

## One note on the comment

The new comment describes the needles rather than spelling them out. This file is walked like
any other, so writing one literally — even in a comment — makes the test report itself. The
first draft of this change did exactly that and failed; the guard caught it.

### DIFF
--- a/compiler/tests/tests/rng_contract.rs
+++ b/compiler/tests/tests/rng_contract.rs
@@ -220,12 +220,23 @@ fn rng_magic_is_owned_by_the_contract() {
     ];
     let stride = ["9e37", "79b9", "7f4a", "7c15"].concat();
     let ambient_mask = ["a5a5", "a5a5"].concat();
+    // Every needle here is a CONSTANT the contract chose. The right-shift width
+    // used to sit in this list and is deliberately absent: it is not a constant
+    // but the shift implied by taking the top 24 bits of a 64-bit hash, which
+    // is what every hash-to-float writes. It also had no detection power to
+    // lose — in `ptir_rng_hash_uniform` that shift sits three lines from the
+    // scale divisor and the golden-ratio stride, so a real transcription trips
+    // those needles anyway. A needle that catches nothing the others miss while
+    // inventing false positives is not a guard rail.
+    //
+    // Needles are assembled from fragments because this file is itself walked:
+    // spelling one out in source — even in a comment — makes the test report
+    // itself. Keep it that way.
     let magic = [
         ["3c79", "ac49", "2ba7", "b653"].concat(),
         ["1c69", "b3f7", "4ac4", "ae35"].concat(),
         stride.clone(),
         ambient_mask.clone(),
-        [">>", "40"].concat(),
         ["16777216", ".0"].concat(),
     ];
 


shsym:
### Removing a needle from a guard rail — here is the evidence, and how to re-run it

This PR deletes an entry from a detector's magic set, which is exactly the kind of change
that should be refused on sight unless the numbers are on the table. They are below, and the
scan is reproducible rather than asserted.

**Result: the dropped needle had one false positive and zero true positives.**

| | offenders repo-wide |
| --- | --- |
| needle set including the right-shift width | **1** — `driver/metal/tests/llama_numerics_test.cpp` |
| with the shift removed (this PR) | **0** |

That single offender is `hashf` at `llama_numerics_test.cpp:124`: MurmurHash3 finalizer
multipliers, mapping to `[-1, 1)` with a different divisor. It shares **no constant** with the
contract — of the six needles it matches only the shift width, and the shift is not a constant
anyone chose, it is what taking the top 24 bits of a 64-bit hash forces you to write.

Nothing stops being detected. A real transcription of `ptir_rng_hash_uniform` still trips the
scale divisor and the golden-ratio stride, which are needles in their own right and sit three
lines from the shift in every generated home:

| | mult1 | mult2 | stride | mask | shift | divisor |
| --- | --- | --- | --- | --- | --- | --- |
| `compiler/ir/src/rng.rs` | Y | Y | Y | Y | – | – |
| `compiler/codegen/include/rng_contract.generated.h` | Y | Y | Y | Y | Y | Y |
| `compiler/codegen/include/ptir_rng.generated.metal` | Y | Y | Y | Y | Y | Y |
| `driver/metal/src/kernels/ptir_rng.generated.metal` | Y | Y | Y | Y | Y | Y |
| `driver/metal/tests/llama_numerics_test.cpp` | – | – | – | – | **Y** | – |

The shift never appears alone in a genuine transcription, and appears alone only in the false
positive.

### Re-run it yourself

The guard `panic!`s on the first offender it finds, so it cannot tell you whether a fix
uncovers the next one. This runs the same walk outside the test and reports the whole set,
under both needle sets. Save as `scan.py` and run `python3 scan.py .` from the repo root:

<details><summary><code>scan.py</code></summary>

```python
#!/usr/bin/env python3
"""Re-run rng_contract's ownership scan outside the test, both needle sets.

  python3 scan.py /path/to/repo

Mirrors visit_sources (same skip-dirs, same extensions) and the same
normalization: lowercase, underscores and whitespace stripped. Needles are
assembled from fragments for the reason the test does it -- this file would
otherwise match itself.
"""
import os, sys

ROOT = sys.argv[1] if len(sys.argv) > 1 else "."
SKIP = {".git", ".claude", "target", "build", "_deps", "node_modules", "third_party", "vendor"}
EXT = {"rs", "c", "cc", "cpp", "cu", "cuh", "h", "hpp", "metal"}

MULT1  = "".join(["3c79", "ac49", "2ba7", "b653"])
MULT2  = "".join(["1c69", "b3f7", "4ac4", "ae35"])
STRIDE = "".join(["9e37", "79b9", "7f4a", "7c15"])
MASK   = "".join(["a5a5", "a5a5"])
DIV    = "".join(["16777216", ".0"])
SHIFT  = "".join([">>", "40"])          # the needle this change removes

NAMES = {MULT1: "splitmix mult 1", MULT2: "splitmix mult 2", STRIDE: "golden-ratio stride",
         MASK: "ambient mask", DIV: "scale divisor", SHIFT: "right-shift width"}

OWNERS = {"compiler/ir/src/rng.rs",
          "compiler/codegen/include/rng_contract.generated.h",
          "compiler/codegen/include/ptir_rng.generated.metal",
          "driver/metal/src/kernels/ptir_rng.generated.metal"}
STRIDE_OK = {"driver/cuda/src/batch/forward_graph.hpp", "driver/cuda/src/loader/weight_store_codec.hpp",
             "driver/cuda/src/pipeline/program_identity.hpp", "gateway/src/route.rs",
             "runtime/engine/src/inferlet/linker.rs", "runtime/engine/src/pipeline/offload.rs",
             "driver/cuda/src/ops/gemm.cpp", "driver/cuda/src/ops/tuning_cache.hpp"}
MASK_OK = {"driver/cuda/tests/ptir_tier0_test.cu", "runtime/grammar/src/brle.rs"}

norm = lambda s: "".join(c for c in s if c != "_" and not c.isspace()).lower()

files, hits = 0, {n: [] for n in NAMES}
for dirpath, dirnames, filenames in os.walk(ROOT):
    dirnames[:] = [d for d in dirnames if d not in SKIP]
    for fn in filenames:
        if fn.rsplit(".", 1)[-1] not in EXT or "." not in fn:
            continue
        rel = os.path.relpath(os.path.join(dirpath, fn), ROOT)
        files += 1
        try:
            text = norm(open(os.path.join(dirpath, fn), encoding="utf-8").read())
        except Exception:
            continue
        for n in NAMES:
            if n in text:
                hits[n].append(rel)

def offenders(needles):
    out = []
    for n in needles:
        for f in sorted(hits[n]):
            if f in OWNERS: continue
            if n is STRIDE and f in STRIDE_OK: continue
            if n is MASK and f in MASK_OK: continue
            out.append((f, NAMES[n]))
    return out

print(f"scanned {files} source files under {os.path.abspath(ROOT)}\n")
print("per-needle: files matching, and how many are declared owners")
for n in [MULT1, MULT2, STRIDE, MASK, DIV, SHIFT]:
    owned = sum(1 for f in hits[n] if f in OWNERS)
    print(f"  {NAMES[n]:<20} {len(hits[n])} matching, {owned} owners")

before = offenders([MULT1, MULT2, STRIDE, MASK, DIV, SHIFT])
after  = offenders([MULT1, MULT2, STRIDE, MASK, DIV])
print(f"\nBEFORE (needle set including the shift): {len(before)} offender(s)")
for f, n in before: print(f"    {f}   <- {n}")
print(f"\nAFTER  (shift removed):                  {len(after)} offender(s)")
for f, n in after: print(f"    {f}   <- {n}")
```
</details>

Output on a clean checkout of this branch:

```
scanned 1285 source files

per-needle: files matching, and how many are declared owners
  splitmix mult 1      3 matching, 3 owners
  splitmix mult 2      3 matching, 3 owners
  golden-ratio stride  8 matching, 3 owners
  ambient mask         5 matching, 3 owners
  scale divisor        2 matching, 2 owners
  right-shift width    3 matching, 2 owners

BEFORE (needle set including the shift): 1 offender(s)
    driver/metal/tests/llama_numerics_test.cpp   <- right-shift width

AFTER  (shift removed):                  0 offender(s)
```

**Correction, so the numbers match what you will see.** This block first read `1286` files,
with every owner count one higher. That run was done in a tree where the project had already
been built, and `driver/metal/src/kernels/ptir_rng.generated.metal` — one of the four declared
owners — is **untracked**: CMake stages it at configure time, so it exists only after a build.
A clean checkout, and CI, see 1285 files and three of the four owners on disk. If you have
built the project, you will get 1286 and the higher owner counts; both are correct for their
tree.

**The offender counts are identical either way: 1 before the change, 0 after.** The conclusion
does not move — the missing file is a declared owner, and owners are exempt from the check, so
its presence or absence cannot add or remove an offender.

Note the scale divisor line: it matches only declared owners, in either tree. That needle is
what actually pins the uniform conversion, and it has no false positives at all.

### Why the comment does not spell the constant out

A reasonable question, since the new comment talks around the needles instead of naming them.
`rng_contract.rs` is walked like every other source file, so writing a needle literally —
**even inside a comment** — makes the test report itself. The first draft of this change did
exactly that and failed with
`PTIR RNG magic ... is independently transcribed in compiler/tests/tests/rng_contract.rs`.
That is why every needle in the file is assembled from fragments, and why the added comment
keeps to that idiom. There is a line saying so, so the next person does not rediscover it.

### Not done, deliberately

`owners` is unchanged — adding the Metal test there would declare it a home for the contract
and legitimize the duplication the test exists to prevent. `llama_numerics_test.cpp` is
untouched — bending correct code to satisfy a bad check. And this is not the
`unrelated_*_users` case: those lists are for needles that *do* carry detection power and also
collide coincidentally (the golden-ratio word in a splitmix id generator, for instance).
Exempting one file would keep a needle with no upside and leave the same trap for the next
hash-to-float anyone writes.

shsym:
Rebased onto `dev-sslee`; carries exactly one commit touching one file, `compiler/tests/tests/rng_contract.rs`.

Re-measured at the branch base in a clean checkout: `rng_magic_is_owned_by_the_contract` still fails there on the same single false positive, so this is not a no-op against its real target.

Marked ready for review.

shsym:

PR #527, opened 2026-08-11
docs(website): sync the pie run flag table with the actual CLI

aarushkandukoori:
Small docs fix to the `pie run` reference.

**What**: The options table in `website/docs/reference/pie.mdx` documented a `--log <PATH>` flag that `pie run` does not accept — there is no such field on `RunArgs`, so clap rejects it as an unexpected argument. Meanwhile four real flags were missing from the table. I removed the phantom row and added:

- `--input <JSON>` — inferlet input as a JSON string (default `{}`), mutually exclusive with the trailing `--` args
- `--stdout` — stream inferlet stdout/stderr as produced; otherwise only the final return value prints
- `--debug` — engine/driver/server/inferlet diagnostics
- `-q`, `--quiet` — suppress progress chatter on stderr

**How verified**: Each row is taken from the `RunArgs` doc comments in `server/src/cli/run_cmd.rs` (lines 48-71), and the `--input` mutual-exclusion note matches the explicit bail at run_cmd.rs:82-85. Grepping `server/src/cli/` confirms no `--log` flag exists anywhere in the CLI.

I deliberately kept this to the `run` table. While checking, I noticed a few other flags the reference also omits — `--force` on `config init`, `--all` on `model download`, `-y`/`--yes` on `model remove`, and the `pie inferlet` subcommand in the top-level list. Happy to fold those into this PR or a follow-up if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated the `pie run` CLI reference with new `--input`, `--stdout`, `--debug`, and `--quiet` options.
  * Removed documentation for the deprecated `--log <PATH>` option.

### DIFF
--- a/website/docs/reference/pie.mdx
+++ b/website/docs/reference/pie.mdx
@@ -87,7 +87,10 @@ pie run [OPTIONS] [INFERLET] [-- ARGS...]
 | `-m`, `--manifest <PATH>` | Path to the inferlet's `Pie.toml` manifest. **Required** with `--path`. |
 | `-c`, `--config <PATH>` | Path to a TOML config file. |
 | `--port <PORT>` | Override `[server].port`. |
-| `--log <PATH>` | Write logs to this file. |
+| `--input <JSON>` | Inferlet input as a JSON string (default: `{}`). Mutually exclusive with the trailing `--` args. |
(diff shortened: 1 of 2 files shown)


ingim:
Thank you @aarushkandukoori !

ingim:

PR #530, opened 2026-08-15
gemm: let a run ask for a dense GEMM tactic that is not a race

shsym:
`DenseGemmTuner` picks each bf16 GEMM's kernel by racing the candidates on the
wall clock and taking the first one within `kGemmTacticMargin` (0.98) of the
fastest, in candidate order. Two percent is jitter on a shared GPU, so the
winner is decided by what else the device was doing during the probe. A
different kernel is a different order of accumulation over K, and the model's
last GEMM before the argmax is inside the tuned range at every serving shape,
so a greedy decode can silently take a different token from one run to the
next.

Three properties make that stick rather than average out:

- the verdict is written to `dense_gemm.txt` under the pod's `PIE_HOME` and
  replayed forever, so one noisy measurement is frozen into every later run on
  that machine;
- `dense_key` hashes **M**, the fire's token count, so a shape's tactic depends
  on who else was in the batch the first time it was measured;
- a shape outside graph capture must be `seen` twice before it is tuned, so its
  first occurrence runs different arithmetic from its second.

This is the same failure the in-place split-K ban ~470 lines up in the same
file is written against — the one that flipped GLM-5.2's step-13 argmax about
half the time through the LM head's reduction order. That fix pinned the
reduction scheme *inside* a kernel and left the choice *of* kernel racing. And
there was no way to turn the race off: the only `getenv` in the file is
`PIE_GEMM_TUNE_LOG`, which only logs.

## The change

`PIE_GEMM_DETERMINISTIC=1` makes `dense_tactic_for` decline every shape.
Nothing is measured, nothing is read back from the disk cache, and nothing is
written to it. The caller falls to the shape ladder at the bottom of
`gemm_bf16_impl` — the GEMV at M=1, cuBLASLt's shape-derived algorithm above
the `min_m`/`min_n`/`min_k` thresholds, `cublasGemmEx` otherwise — which is
what this file did before the tuner existed and is a pure function of
(M, N, K, beta).

Unset, nothing changes. The addition on the default path is one predictable
branch on a cached `static bool`, in front of work that already includes a
`cudaMalloc` and a stream sync.

Cost when it *is* set: the tuner's measured wins are given up, so a
deterministic run is a pre-tuner-speed run. That is the intended trade for a
parity or bisect run and is stated in the header.

## Why `dense_key` still hashes M

Considered and rejected. Dropping M would not buy batch-invariance and would
actively corrupt the cache: a tactic's `algo` is an index into
`Bf16LtPlan::heuristics`, and `build_lt_plan`/`Bf16LtKey` build that plan per
(M, N, K) — index 3 at M=1 and index 3 at M=8 name different algorithms, so a
key shared across M replays whichever one happened to be measured first.
Bucketing M has the same defect at every bucket boundary.

M is a real axis of the arithmetic in any case: an M=8 GEMM tiles its K
reduction differently from an M=1 GEMV whatever tactic either is handed. Logits
that do not move with batch composition are a strictly larger change than this
one, and this PR does not claim it.

## Evidence, and what is *not* evidence

**Reasoned from the code, not measured.** The determinism claim here is an
argument about the code — that with the tuner declined, the surviving path is a
function of the shape alone — and no token stream was compared to check it.
This is a Mac; the build host has no GPU.

Verified:

- `tests/gemm_determinism_test.cpp` compiles clean under
  `clang++ -std=c++20 -Wall -Wextra` and passes, 6 cases.
- It is non-vacuous: replacing the reader's body with a hardcoded `false`
  turns it red on all three truthy spellings, with the message naming the
  cause. The pure-parser assertions stay green through that mutation, which is
  exactly why the test re-execs instead of only parsing.

Not verified:

- `driver/cuda/src/ops/gemm.cpp` is **uncompiled**. The change there is one
  `#include` and one `if`, but no nvcc has seen it. Please run the CUDA build
  (or apply the `ci-cuda` label) before merging.
- Nothing was run on a GPU.

**The experiment that would confirm the claim:** same prompt, same pod, greedy
decode, `PIE_GEMM_DETERMINISTIC=1`. Delete `dense_gemm.txt`, run once, run
again against the now-warm cache, and diff the token ids. They should be
identical, and identical to a third run from a cold cache under load. Without
the switch the same procedure is the reproduction: run cold and warm and expect
the streams to diverge at the first near-tie.

## Also noticed, not touched

`dense_tactic_already_gemv` (gemm.cpp:~888) has no callers anywhere in the
tree. Pre-existing and unrelated; left alone.

## Scope of the guarantee (added after cross-machine testing)

`PIE_GEMM_DETERMINISTIC=1` delivers **boot-invariance on one machine**: the
default path is a pure function of (M, N, K, beta), so two runs on the same
pod produce the same arithmetic. It does **not** claim machine-invariance.
A separately observed cross-machine divergence (bimodal outputs across four
same-image, same-config pods; cause under investigation) enters at a layer
whose GEMMs are already on the default path — if that lands inside
cuBLAS/cuBLASLt kernel dispatch, this mode will not cure it, and curing it
would need a pinned-ALGORITHM mode rather than a no-tuner mode. Cite this
flag as "same machine, any boot", never as "any machine".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -410,9 +410,10 @@ jobs:
             -path '*/out/cuda/build' -print -quit)"
           test -n "$CUDA_BUILD_DIR"
           cmake --build "$CUDA_BUILD_DIR" \
-            --target test_entry_validation test_pipeline_dispatch_bind
+            --target test_entry_validation test_pipeline_dispatch_bind \
+                    test_gemm_determinism
           ctest --test-dir "$CUDA_BUILD_DIR" --output-on-failure \
-            -R '^(entry_validation|pipeline_dispatch_bind)$'
+            -R '^(entry_validation|pipeline_dispatch_bind|gemm_determinism)$'
 
       - name: Verify no CUDA shared deps
         run: |
--- a/driver/cuda/CMakeLists.txt
+++ b/driver/cuda/CMakeLists.txt
@@ -1231,6 +1231,11 @@ pie_add_cuda_target(test_flashinfer_decode_head_dim_guard
   SOURCES tests/test_flashinfer_decode_head_dim_guard.cpp
   LIBS pie_driver_cuda_lib)
 pie_add_cuda_target(test_graph_variant SOURCES tests/graph_variant_test.cpp)
+# PIE_GEMM_DETERMINISTIC is still read from the environment. Names no src/ file
+# because the switch is header-only, and no device because it is host C++; it
+# re-execs itself, so it must stay a standalone executable.
+pie_add_cuda_target(test_gemm_determinism
+  SOURCES tests/gemm_determinism_test.cpp)
 # Reaches `pie/driver/fire/descriptor.hpp`, which pulls in `launch/program.hpp`
 # and so `pie_driver_abi.h` — the ABI include dir is not optional here even
 # though nothing in the test names it directly.
--- a/driver/cuda/src/ops/gemm.cpp
+++ b/driver/cuda/src/ops/gemm.cpp
@@ -24,6 +24,7 @@
 #include "kernels/gemv.hpp"
 #include "kernels/quant_bf16_to_fp8.hpp"
 #include "kernels/residual_add.hpp"
+#include "ops/gemm_determinism.hpp"
 #include "ops/tuning_cache.hpp"
 
 #ifdef PIE_CUDA_HAS_MARLIN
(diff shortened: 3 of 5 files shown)


shsym:
Cross-referencing independent work on the same defect, found after this PR was opened.

A parallel investigation on the Qwen3.6-27B-FP8 parity ticket reached the same root cause
by measurement rather than by reading, and its diagnosis is sharper than this PR's. Using
per-tensor activation dumps it localised the divergence to `L03_k_gemm` — the K projection
— with bit-identical inputs, weights, buffer addresses and dequant-cache fates. The trigger
is not prefix identity but a **same-M predecessor**: `dense_tactic_for` makes a shape prove
it recurs, so the first call of a `(M,N,K,beta)` runs the default cuBLAS path and the second
onward runs the tuned tactic. At `M=193,N=1024,K=5120` that is cuBLASLt algo 5 versus
GemmEx, measured at **484 ULP flips in K**, amplifying over 64 layers. GDN layers stayed
exact because their shapes tune to a tactic numerically identical to the default.

Their fix — adopt the tactic at FIRST sighting — is `fix/dense-tuner-first-sight`
(`ba35d0e1`) on `shsym/pie`, and it is verified bit-exact across all 210 dumped tensors.
It is currently on merge hold.

**That fix should land ahead of this one, and it may make this one unnecessary.** It closes
the discontinuity that actually bit, and it does so with measured evidence; this PR's
determinism claim is reasoned from the code and has not been compiled, let alone run.

The two are not the same change, and the residual case is worth deciding deliberately rather
than by whichever merges first. First-sight adoption makes first and later calls agree within
one engine lifetime. It does not remove the wall-clock race itself, so two pods that both
tune cold can still adopt different tactics — which matters for a cross-run or cross-pod
comparison, and is what the switch here addresses. If that residual is judged acceptable,
close this; if not, this becomes a small addition on top of theirs rather than an alternative
to it.

Either way the verified change should be the one that carries.

shsym:

---

To the lane that owns this PR: the dense-GEMM tactic-cache first-sight fix (ba35d0e15, merged to dev-sslee via PR #535) landed as the accepted determinism remedy, and project memory records this PR as having yielded to it. Is this run-level tactic knob still wanted on its own merits (e.g. as a debug A/B arm — if so it needs a PIE_DEBUG_ name per operator rule and a rebase onto current dev-sslee), or should it close as superseded? Board-level coordination is trying to drain stale open PRs; silence for a day will be read as superseded, but the close will be yours or the operator's, not mine.

PR #531, opened 2026-08-16
engine: one-shot channel poison + FP8-GDN scale-factor load fix

shsym:
Two lane fixes from #2017 debugging.

**1. Deliver channel poison once, then recover.** A failed fire poisons its pass's host-reader channels so the parked reader observes the error. The poison was sticky for the life of the channel, and decoder sessions keep channels across requests — so one admission-time rejection (a frame over the physical budget ceiling, which never touched the device) permanently poisoned the session: every later request failed with 'channel is poisoned' until a restage. Host-side poison is now one-shot: take delivers the error and clears it, read peeks it, first-poison-wins unchanged, driver-published poison epochs (real device faults) stay sticky. All 426 pie-engine lib tests pass; the poison unit test now asserts recovery after delivery.

**2. Ship FP8-GDN scale factors at their checkpoint dtype; widen in the engine** (cherry-pick 0452d7a5 from the score lane). The real Qwen3.6-27B-FP8 load at tp2 threw 'unsupported TileMap Cast u8 -> fp32' because the plan cast gathered weight_scale_inv bands to F32 through an anonymous staged buffer the executor types as u8. Factors now ship at checkpoint dtype (BF16) and the engine widens them with the existing cast kernel. The C++ hunk is the same code already running in the score/f9a62b81-rope-fix binary that produced the graded parity results; family_contracts tests (35) pass on this branch.

### DIFF
--- a/driver/cuda/src/loader/transcode_engine.hpp
+++ b/driver/cuda/src/loader/transcode_engine.hpp
@@ -413,16 +413,31 @@ private:
         }
         // MXFP4 pairs E2M1 elements with E8M0 exponents; Int4B8 pairs
         // biased nibbles with plain BF16 factors; FP8 pairs a byte per element
-        // with F32 reciprocal scales. No kernel reads another's factor format.
+        // with reciprocal scales the kernel reads as F32 — shipped either F32
+        // or BF16 (Qwen3.6-FP8 stores `weight_scale_inv` BF16), and a BF16
+        // slab is widened here with the cast kernel rather than in the plan,
+        // where the cast's staged operand would be untyped bytes. No kernel
+        // reads another's factor format.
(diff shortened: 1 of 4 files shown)


shsym:

---

CI triage before merge: the 4 failing checks (cargo check/test workspace, workspace verify, wit drift, driver-metal) all reproduce byte-for-byte at the lane base aee1b99a — verified locally by running the failing rng_contract test at the base commit (fails identically: 'PTIR RNG magic >>40 is independently transcribed in driver/metal/tests/llama_numerics_test.cpp'). None of them touch this PR's changed crates; pie-engine (426 tests), family_contracts (35), and driver-dummy worker jobs on both OSes are green. Merging on the operator's directive; the inherited lane CI debt is tracked separately.

PR #533, opened 2026-08-16
Converge the parity lineage: qwen3 template fixes and the act-dump steps knob

shsym:
Brings the two remaining parity fixes into the pinned tts-arena lineage so one revision carries every front's work: the five qwen3 chat-template parity commits (render reasoning blocks, tool declarations, and replayed tool calls exactly where the checkpoint's template renders them, fixing the 9B instruct arm's prompt divergence) and the PIE_ACT_DUMP_STEPS knob that act_dump's documentation names but the code never read. Both merge clean and neither touches the hybrid decode paths. The FP8-GDN load and channel-poison fixes are already on this base branch.


shsym:
Wrong base: the tts-arena v0.4 lineage is parked; convergence goes into dev-sslee (v0.5 pre-release) instead. Replacement PR targets dev-sslee with a superset of this content.

PR #536, opened 2026-08-16
Reproduce vLLM's bf16 RoPE table behind a knob, default off

shsym:
## What

An opt-in RoPE path for partial-rotary models that reproduces vLLM's cos/sin-table numerics, behind `PIE_ROPE_VLLM_TABLE=1`. **Default off; this PR does not flip it.**

The table is built **on the host** and uploaded, as vLLM builds `cos_sin_cache` at init and indexes it thereafter. No GPU kernel evaluates trig on this path.

## Why

`rope_theta = 10000000`, `partial_rotary_factor = 0.25`, `head_dim = 256` → `rotary_dim = 64`, making `inv_freq[0]` **exactly 1.0f**: lane 0's angle *in radians is the token position*.

**`__sincosf` does not collapse at long positions** — measured at lane 0: 1.78e-05 (0-200) → 4.24e-03 (past 20000), against a bf16 step of 7.8e-03. Monotonic, bounded. The fix does not work because the intrinsic is inaccurate. It works because **vLLM's values are a specific bf16-rounded table, and only reproducing that table reproduces its bits.**

vLLM computes trig once on the host in fp32, then `cache.to(dtype)` rounds the whole table to bf16; `triton_mrope` loads those bf16 values with no fp32 cast and casts q/k *down*, so the rotate is bf16 end to end. **An implementation that computed trig more accurately would move away from parity, not toward it** — a theme that recurred three times below.

## Where the numerics landed

Device-side trig was measured on real sm_89 against the reference table: **4 mismatches / 1,920,064**, each exactly ±1 bf16 ulp, each within 0.71 fp32 ulp of a rounding midpoint. The rounding-rule hypothesis is dead — `__float2bfloat16` and torch's `.to(bfloat16)` are both round-to-nearest-even, so a tie breaks identically regardless of how many ties exist. Only running the reference's *evaluation* gets the reference's bits, which is why the build moved to the host.

**The reference's trig is Intel MKL VML, not SLEEF.** `ATen/cpu/vml.h` specialises `vcos`/`vsin` onto MKL whenever `AT_MKL_ENABLED() && !__APPLE__`, so `Vectorized<float>::cos()` is never reached on the x86 torch build the reference runs. An earlier arm64 measurement only saw SLEEF because that build had `USE_MKL=OFF`.

Host-side residual against MKL: **19 entries over 0..262143**, only `pos=13852` inside the campaign window. Closing it bit-exactly would require vendoring closed-source, x86-only oneMKL into a CUDA driver's host table builder. **Not proposed.** The open question is whether any of this flips a token, which needs an end-to-end run, not more numerics.

### Host trig backend

`exact` (correctly rounded) is the default, `PIE_ROPE_VLLM_TABLE_TRIG=libm` selects `cosf`/`sinf`. Two independent legs:

1. **Determinism.** Correct rounding has one answer, so the table depends on nothing. `libm` would make Pie's table a property of the build host. *True in principle; not demonstrated by the two glibcs tested — they agree exactly.*
2. **Proximity.** Fraction of angles whose fp32 value differs from the reference: **exact 5.205%, libm 5.297%.** `exact` is also the closer one.

## Design notes that are not incidental

- Table cache keyed by **device** as well as `(theta, rotary_dim)` — TP runs every rank in one process on its own device.
- Upload on a **private stream**: the caller's stream may be under CUDA-graph capture, where a copy is *recorded*, not performed.
- Past capacity the kernel degrades to device trig and bumps an `oob` counter, asserted zero by the test, so parity can't pass while silently on the fallback.
- **Capacity is the known weak point.** Default 262144 (33.5 MB at `rotary_dim=64`), env-overridable. Should come from `cfg.max_position_embeddings` (in scope at all 8 call sites) — deferred to the default-flip change. Gemma-4's per-layer `rope_theta` means one table per distinct theta.

## Knob-off is bit-unchanged, as measured fact

The new behaviour is a separate kernel, and **all 9 pre-existing kernels in `rope.cu` compile to byte-identical SASS at sm_89** before and after (cuobjdump, anon-namespace hash normalized).

## Scope — which `__sincosf` sites are covered

Covered: `rope_partial_bf16_kernel` via its two launchers (6 Qwen3.5 sites; Gemma-4's 2 also route here).

**A correction to the plan this was written from:** `rope_device.cuh`'s `rope_cos_sin` is documented as "what every rope variant has to reproduce exactly" and was expected to be the point of leverage with ~18 sites downstream. It has exactly **two** callers, both in `mla_paged.cu` (DeepSeek MLA); `rope_partial_bf16_kernel` inlines its own `powf`/`__sincosf`. Editing it would have changed DeepSeek and left Qwen3.5 alone.

Not covered (16 sites): `rope_bf16_kernel`, both YaRN kernels, three `qk_rmsnorm_*`, `rope_partial_last_bf16_kernel`, `rope_standard_table_kernel`, `mla_paged.cu`, `split_packed.cu` (3), `dsa_indexer.cu`.

## Tests

`rope_partial_parity.cu` guards **two properties against two references**, and the header says so, because conflating them broke this file repeatedly:

- **default path → STRUCTURE**, vs the HF definition in fp64, with a tolerance.
- **`..._vllm_table` → BITS**, vs vLLM's `cos_sin_cache`. No tolerance; no transcendental evaluated in the test.

The kernel exposes no table, so it is **read back exactly**: drive a head with `a = 1.0, b = 0.0`, giving `out[j] = cos[j]`, `out[j+angles] = sin[j]`. Every operand is bf16 and every product exact, so the output *is* the table row.

The fixture is generated by `gen_rope_vllm_golden.py` from measured inputs — 42 positions / 2688 entries, 19 reference overrides, surveyed to 262143, labelled `x86_64-mkl-vml-ha` and printed on every run.

### Three failure modes this file actually had

1. **Accuracy vs fp64.** An earlier revision demanded the table path be *ten times more accurate* than the default. Contradicts the change's own thesis; silicon confirmed it fails. fp64 truth is now gone from the vLLM checks entirely.
2. **A vacuous fixture, twice.** v1 sampled 13 positions and passed **832/832 with zero tolerance — vacuously**: not one entry lay within even 3 fp32 ulp of a midpoint, and a bf16 table can only disagree near a midpoint. v2 was built around the device-era positions; the host-side fix cleared all four and moved the residual elsewhere, so it went blind again one iteration later. The positions are now an *input*, and the test asserts the slice has teeth **before any GPU work**.
3. **A circular fixture.** The generator built every entry with *Pie's own* construction and called it vLLM's. Harmless where the two agree — which was every sampled entry, which is why it went unnoticed — but at a divergent entry it would have compared Pie to Pie and reported zero. `--override` now carries the reference's *measured* bits, and the generator refuses an override matching what it already produces, and refuses any position past `--survey-max`.

Mismatches are classified `at_override` vs `elsewhere` as a diagnostic. **The bar is unchanged at `total == 0`** — classification is a label, not an allowance.

## Verification

**Done.** Compiles and links for sm_89, CUDA 12.9.86. SASS identity measured. An earlier revision's kernel ran on real sm_89 and produced the 4-mismatch data; the skip guard was proven both directions (GPU → exit 0; `CUDA_VISIBLE_DEVICES=""` → exit 77). Host construction verified against the fixture under macOS libm and Linux glibc 2.35/2.41.

**Not done.** The current test has never executed — no GPU reachable from this environment at any point. Expected on sm_89: parity red at 19, only `pos=13852` in the campaign window.

Deployment base verified rather than assumed: `backend/docker/eval-worker-base.Dockerfile` pins `nvidia/cuda:12.9.1-cudnn-runtime-ubuntu22.04`, i.e. glibc 2.35, which is where every host number above was measured.

## Residual expected even if this is right

`rotary_dim = 64` of `head_dim = 256`, and `layer_types` puts `full_attention` at only 8 of 32 layers — a quarter of the dims in a quarter of the layers. A residual gap should be expected rather than treated as a new defect.


shsym:

---

Heads-up for the A/B you have planned: these knobs were renamed in PR #537 (merged, dev-sslee 3d57c4415) per operator rule — debug instruments carry a debug marker. They are now PIE_DEBUG_ROPE_VLLM_TABLE, PIE_DEBUG_ROPE_VLLM_TABLE_TRIG, PIE_DEBUG_ROPE_VLLM_TABLE_MAX_POS. Exporting the old names against a build at or past 3d57c4415 silently does nothing — set the new names, and verify the knob took effect from the engine's own log line rather than assuming.

